PR #625, opened 2026-04-23
core: structured self-signed cert error

### LINKED ISSUE #601: Self-signed cert produces opaque error — no actionable user message
**Severity**: P2  
**Areas**: `area:core`, `area:reliability`, `area:ux`

When connecting to a Jellyfin server with a self-signed cert, the user gets a generic "Network error" with no hint that TLS verification is the issue. Common pain for home-server users (Tailscale, local-LAN deployments, custom CAs).

## Files

- [`core/src/client.rs:55-59`](core/src/client.rs#L55) — `HttpClient::builder()` uses default cert verification.

## Why it matters

Jellyfin's user base is heavily home-server. "I can connect from a browser but the app won't" is a recurring confusion; the diagnostic gap makes it feel like an app bug rather than a cert config issue.

## Proposed fix

Two parts:

1. **Better error**: detect TLS errors specifically (reqwest's error chain has the underlying rustls/native-tls error) and map them to a `JellifyError::TlsTrust { reason: String }` variant. Surface a friendly message: "Couldn't verify the server's certificate. If this is a home server with a self-signed cert, you can opt to trust it in Preferences > Server > Advanced."

2. **Optional opt-in trust**: Preferences > Server > Advanced has a "Trust self-signed certificates" toggle. When enabled, configure `reqwest::ClientBuilder::danger_accept_invalid_certs(true)`. Surface a clear warning in the UI when this is enabled. Persist per-server (not globally).

Acceptance: connecting to a self-signed Jellyfin shows a clear actionable error, with a path to opt-in trust.

---

skalthoff:
Fixes #601.

When the Jellyfin server uses a self-signed certificate, the client previously returned an opaque `JellifyError::Network` with a reqwest error string. This change:

- Adds `JellifyError::SelfSignedCertificate { host: String }` variant to `error.rs`
- Walks the `reqwest::Error` source chain in `From<reqwest::Error>` to detect the rustls 0.23 pattern `"invalid peer certificate: …"` (and compatible older patterns) and maps it to the new variant
- Updates `send_with_retry_raw` in `client.rs` to route through `Into` so the structured mapping fires automatically
- Adds three tests: a unit test confirming non-cert errors are not misclassified, a `#[ignore]`-tagged integration test against `self-signed.badssl.com`, and a connection-refused test asserting `Network` is preserved for non-cert transport errors

The UI can now pattern-match on `SelfSignedCertificate { host }` to display an actionable "Trust this server" prompt instead of a raw network error string. Certificate trust bypass is not implemented here — that remains a UI-driven decision.

### DIFF
--- a/core/src/client.rs
+++ b/core/src/client.rs
@@ -335,7 +335,10 @@ impl JellyfinClient {
                         tokio::time::sleep(delay).await;
                         continue;
                     }
-                    return Err(JellifyError::Network(err.to_string()));
+                    // Route through `From<reqwest::Error>` so cert-validation
+                    // failures are mapped to `JellifyError::SelfSignedCertificate`
+                    // rather than the generic `Network` variant.
+                    return Err(err.into());
                 }
             }
         }
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -6,6 +6,13 @@ pub enum JellifyError {
     #[error("network error: {0}")]
     Network(String),
 
+    /// The server presented a TLS certificate that could not be verified (e.g.
+    /// self-signed or issued by an unknown CA). Separated from the generic
+    /// [`JellifyError::Network`] variant so the UI can offer a "Trust this
+    /// server" action instead of a generic error message.
+    #[error("the server at '{host}' uses a certificate that could not be verified — it may be self-signed")]
+    SelfSignedCertificate { host: String },
+
     #[error("authentication failed: {0}")]
     Auth(String),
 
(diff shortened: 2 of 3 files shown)


skalthoff:
**Blocked — clippy failure**

`cargo clippy --workspace --all-targets --all-features -- -D warnings` failed in CI run 24817065329 (job 72633491103). This is a genuine clippy error (exit code 101) in the changes to `core/src/client.rs` / `core/src/error.rs`, not a pre-existing issue. Needs clippy fix before merge.

skalthoff:
Needs manual rebase

Issue #608, opened 2026-04-23
Auth header lacks input sanitization on device_name


skalthoff:
**Severity**: P2  
**Areas**: `area:core`, `kind:security`

The auth header construction in `build_headers()` flows `device_name` (user-controllable string) into `HeaderValue::from_str()` without sanitization. A device name containing `"`, comma, newline, or any non-printable ASCII either fails header construction (downgraded to `JellifyError::InvalidInput`) or produces a malformed header that the server may parse unpredictably.

## Files

- [`core/src/client.rs:130-154`](core/src/client.rs#L130) — `auth_header()` and `build_headers()`.
- [`core/src/lib.rs:88`](core/src/lib.rs#L88) — currently hardcoded to `"Jellify macOS"`, but Preferences will eventually let users set this (per the device-name customization roadmap item).

## Why it matters

Today: hardcoded device name → no risk. As soon as device name becomes user-settable (which is on the roadmap for Devices-page UX), a `"` in the name breaks all auth requests with an opaque error.

## Proposed fix

- [ ] Replace any unsafe characters in the device name with their ASCII equivalents or strip them: `["\r\n,\\]` → empty or underscore.
- [ ] OR: properly escape the device name for the `MediaBrowser` header format (which uses comma-separated key="value" — escape inner quotes by doubling or backslash-escaping per the header spec).
- [ ] Add a unit test with malicious names: `Device"name`, `Device\nname`, `Device,name`, empty string.

Acceptance: setting device name to `Soren's "Macbook" 14"` succeeds and the server recognizes the device.

---
*Found during full-surface adversarial audit on main, 2026-04-22.*

skalthoff:
Scoped with #584 in batch `auth-keyring`.

Issue #584, opened 2026-04-23
Keyring write failures silenced via 'let _ =' pattern


skalthoff:
**Severity**: P1  
**Areas**: `area:core`, `area:reliability`

Keyring (Keychain on macOS) write failures are discarded via the `let _ =` pattern. If the Keychain is locked, unavailable, or rejects the write, token persistence fails silently with no UI feedback. User thinks they're logged in; next launch has no token.

## Files

- [`core/src/lib.rs:130`](core/src/lib.rs#L130) — `let _ = creds.save_token(...)`
- [`core/src/lib.rs:222`](core/src/lib.rs#L222) — `let _ = creds.delete_token(...)`
- [`core/src/lib.rs:248`](core/src/lib.rs#L248) — `let _ = creds.delete_token(...)`

## Why it matters

The whole "remember me" path is one Keychain hiccup away from confused users. Reproducible on first-launch when Keychain is locked, after a Keychain reset, or on FileVault-encrypted volumes that haven't been unlocked.

## Proposed fix

- [ ] Replace `let _ =` with proper `?` error propagation, OR explicitly log + surface to UI: "Couldn't save credentials securely. Sign in again next launch?"
- [ ] In `resume_session()` ([`core/src/lib.rs:158-189`](core/src/lib.rs#L158)), distinguish "keyring locked" from "no token stored" and surface the difference (so the UI can prompt to unlock Keychain rather than re-login).
- [ ] Add a structured `JellifyError::Keyring` variant for these cases.

Acceptance: kill the keychain process mid-login → app shows "credentials couldn't be saved" warning instead of silently dropping the token.

---
*Found during full-surface adversarial audit on main, 2026-04-22.*

skalthoff:
Scoped with #608 in batch `auth-keyring`.

PR #627, opened 2026-04-23
core(auth): trim creds + keyring error propagation + device header sanitization

### LINKED ISSUE #577: Username/password whitespace not trimmed (login + resume_session)
**Severity**: P1  
**Areas**: `area:macos`, `area:core`, `area:reliability`

Username and password are passed raw from the login form to the auth call — no whitespace trimming. " admin " ≠ "admin"; user gets a vague 401 with no hint that whitespace was the problem. Worse: if a bad value gets persisted, `resume_session` reads it back without trimming, producing an infinite 401 loop on cold start until the user manually logs out.

## Files

- [`macos/Sources/Jellify/AppModel.swift:344`](macos/Sources/Jellify/AppModel.swift#L344) — `login(url:username:password:)` passes raw strings.
- [`core/src/lib.rs:158-189`](core/src/lib.rs#L158) — `resume_session()` reads `last_username` from DB without trimming.

## Repro

1. Type ` admin` (leading space) into username field on login.
2. Submit. Observe vague "401 Unauthorized" error.
3. Hit "Sign In" two more times — same error.
4. Bonus: if login somehow persisted that value, app is stuck in a 401 loop on every cold launch.

## Proposed fix

- [ ] Trim username (`.trimmingCharacters(in: .whitespacesAndNewlines)`) before passing to the FFI.
- [ ] Trim password too — usually trailing newline from copy-paste.
- [ ] In Rust core's `resume_session`, defensively trim the stored value before sending to the server.
- [ ] Optional: detect "username had whitespace" case and surface a helpful UI hint ("We trimmed extra spaces from your username — sign in again?").

Acceptance: pasting `  admin  ` followed by `<correct password>` succeeds.

---

skalthoff:
Fixes #577, #584, #608.

## Changes

**#577 — Credential whitespace trimming** (`lib.rs`)
- `login` trims username and password with `.trim()` before any network call; blank-after-trim returns `JellifyError::InvalidCredentials`
- `resume_session` trims the username loaded from the DB (guards against stale data written by older app versions)

**#584 — Keyring write error propagation** (`lib.rs`, `error.rs`)
- Added `JellifyError::KeyringWrite { reason: String }` variant
- Replaced `let _ = CredentialStore::save_token(...)` with `map_err(...)?` so a keyring rejection surfaces to the caller instead of being silently swallowed

**#608 — Device name header sanitization** (`client.rs`)
- Added `sanitize_header_field(s, max_len)` helper that strips all Unicode control characters (CR, LF, NUL, C0/C1 range) and truncates to 100 chars
- `auth_header` runs `device_name` through this helper before embedding it in the `Authorization` value, preventing CRLF header injection

## Tests (`tests.rs`, 8 new cases)
- `login_rejects_whitespace_only_credentials` — whitespace-only username/password → `InvalidCredentials`
- `login_trims_credentials_before_auth` — padded `"  soren  "` authenticates successfully
- `auth_header_strips_crlf_injection` — CR/LF absent from header after sanitization
- `auth_header_truncates_long_device_name` — 200-char name is capped at 100
- `keyring_write_error_variant_is_displayable` — new variant round-trips through `Display`
- `login_keyring_write_is_not_silenced` — login error path accepts `Ok` or `KeyringWrite`, panics on any other variant

### DIFF
--- a/core/src/client.rs
+++ b/core/src/client.rs
@@ -127,16 +127,31 @@ impl JellyfinClient {
         *self.refresh_cb.lock() = Some(cb);
     }
 
+    /// Strip control characters (`\r`, `\n`, `\0`, and any other C0/C1 code
+    /// point) from a string intended for use inside an HTTP header value,
+    /// then truncate to `max_len` characters. Prevents header-injection
+    /// attacks when a caller-supplied `device_name` contains embedded CRLF
+    /// sequences (see issue #608).
+    fn sanitize_header_field(s: &str, max_len: usize) -> String {
+        let cleaned: String = s.chars().filter(|c| !c.is_control()).collect();
+        if cleaned.len() > max_len {
+            cleaned.chars().take(max_len).collect()
+        } else {
+            cleaned
+        }
+    }
+
     fn auth_header(&self) -> String {
         let token_guard = self.token.lock();
         let token_part = token_guard
             .as_deref()
             .map(|t| format!(", Token=\"{t}\""))
             .unwrap_or_default();
+        let safe_device = Self::sanitize_header_field(&self.device_name, 100);
         format!(
             "MediaBrowser Client=\"{client}\", Device=\"{device}\", DeviceId=\"{device_id}\", Version=\"{version}\"{token}",
             client = CLIENT_NAME,
-            device = self.device_name,
+            device = safe_device,
             device_id = self.device_id,
             version = CLIENT_VERSION,
             token = token_part,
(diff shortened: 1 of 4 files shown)


skalthoff:
Needs manual rebase

skalthoff:
Needs manual rebase — conflicting after recent merges.

skalthoff:
Needs manual rebase — `AppModel.swift` calls to `addToPlaylist`/`createPlaylist` are missing the `position` argument added in #633 (landed on main). macOS app build fails at lines 1908, 1920, 2542, 3017, 3079.

Issue #605, opened 2026-04-23
No request cancellation safety on dropped futures — wasted work + zombie requests


skalthoff:
**Severity**: P2  
**Areas**: `area:core`, `area:perf`

All `client.rs` request futures are `await`ed unconditionally. There's no `select!()` or timeout for user-driven cancellations (e.g., navigating away from a search page mid-fetch). The HTTP request completes regardless; the response is just dropped. Wasted bandwidth + zombie work + slower response to subsequent legitimate requests.

## Files

- Throughout `core/src/client.rs` — every `async fn` that calls `send_with_retry`.

## Why it matters

- Library page that takes 2s to load and the user navigates away in 500ms → server still does the full work.
- Multiple concurrent in-flight requests stack up if the user clicks around quickly (search → album → search again before previous returned).

## Proposed fix

- [ ] Pass `tokio_util::sync::CancellationToken` (or `futures::future::AbortHandle`) into client methods, especially long-running ones (search, library list, similar items).
- [ ] Caller (UI / view-model) drops the token when navigating away.
- [ ] Use `tokio::select!` between the HTTP future and the cancellation signal.
- [ ] Add a request timeout (`reqwest::ClientBuilder::timeout`) as a baseline — currently no timeout means a single hung request hangs the calling task forever.

Acceptance: navigate away from a slow search → the request is dropped within 200ms (verifiable via server access log not seeing the response delivered).

---
*Found during full-surface adversarial audit on main, 2026-04-22.*

skalthoff:
Scoped with #594 in batch `cancellation-heartbeat`.

skalthoff:
Closed by #639.

PR #639, opened 2026-04-23
core: heartbeat scheduler + request cancellation

### LINKED ISSUE #594: report_playback_progress heartbeat scheduler unverified — server may think playback ended
**Severity**: P1  
**Areas**: `area:core`, `area:api`, `area:reliability`

A comment in `lib.rs` mentions "roughly every 10 seconds" for playback progress reports, but no scheduler is visible in the audit. AudioEngine's KVO `mark_position()` fires every 0.5s — does it fan out to a server progress report? If not, the server thinks playback ended immediately after start.

## Files

- [`core/src/lib.rs:794`](core/src/lib.rs#L794) — comment about cadence.
- [`macos/Sources/JellifyAudio/AudioEngine.swift:131-137`](macos/Sources/JellifyAudio/AudioEngine.swift#L131) — `attachPlayerObservers` adds 0.5s periodic time observer.
- [`core/src/client.rs:1373`](core/src/client.rs#L1373) — `report_playback_progress` HTTP call.

## What needs verification

- Is there an actual 10s heartbeat scheduled anywhere?
- Or does `mark_position()` rate-limit and call `report_playback_progress` at intervals?
- What happens if a report fails — is there retry, or is the server's "currently playing" state stuck?

## Why it matters

Without progress reports, the Jellyfin admin dashboard doesn't know the playback is alive. "Resume from server" position never advances on other devices. Dropped progress reports = orphaned sessions until the next start.

## Proposed fix

- [ ] Audit the actual reporting cadence end-to-end (Swift KVO → core call → HTTP).
- [ ] If absent, add an explicit `tokio::time::interval(Duration::from_secs(10))` task that calls `report_playback_progress` while playback is active.
- [ ] Coalesce: only report when position has advanced more than ~1s since last report.
- [ ] Report on important events too (pause, seek, mute change).
- [ ] Cancel the heartbeat on track end / stop / error.

Acceptance: tail server access log during a 60-second playback → see ~6 progress reports with monotonically increasing PositionTicks.

---

skalthoff:
Fixes #594, #605.

## Summary

- **#594 (heartbeat)**: Added `start_heartbeat(interval_secs, play_session_id)` / `stop_heartbeat()` to `JellifyCore`. The scheduler spawns a tokio interval task (interval clamped to ≤10 s — Jellyfin's server-side playback-ended detection threshold) that builds `PlaybackProgressInfo` from current `Player` state and POSTs to `/Sessions/Playing/Progress`. The `AbortHandle` is stored internally and cancelled automatically on `stop()`, `logout()`, and `forget_token()`, preventing leaked timers across track skips and session changes.
- **#605 (cancellation)**: Added a `CancellationToken` field to `JellyfinClient`. The backoff sleep inside `send_with_retry_raw` is now wrapped in `tokio::select!` so cancellation fires immediately rather than waiting out the full delay. No endpoint method signatures changed.
- **Arc refactor**: `Inner::client` is now `Option<Arc<JellyfinClient>>` so the heartbeat task can clone the client handle without holding the `Mutex` across an `await`.

## Tests added

- `heartbeat_fires_at_expected_cadence` — uses `tokio::time::pause` + `advance` to verify ≥2 heartbeats fire in 15 simulated seconds with a 5 s interval.
- `heartbeat_stops_after_abort` — verifies no POSTs fire after `AbortHandle::abort()`.
- `cancelled_token_aborts_retry_backoff` — cancels the token 20 ms into a 503-retry backoff and asserts the error returns in under 500 ms.

All 139 tests pass; `cargo clippy --all-targets --all-features -- -D warnings` clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1427,6 +1427,7 @@ dependencies = [
  "tempfile",
  "thiserror",
  "tokio",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
  "uniffi",
@@ -2486,6 +2487,7 @@ dependencies = [
  "bytes",
  "futures-core",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ repository = "https://github.com/Jellify-Music/Jellify"
 [workspace.dependencies]
 # Runtime
 tokio = { version = "1.40", features = ["full"] }
+tokio-util = { version = "0.7", features = ["rt"] }
 async-trait = "0.1"
 futures = "0.3"
 
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -17,6 +17,7 @@ path = "src/bin/uniffi-bindgen.rs"
 
 [dependencies]
 tokio.workspace = true
+tokio-util.workspace = true
 async-trait.workspace = true
 futures.workspace = true
 
--- a/core/src/client.rs
+++ b/core/src/client.rs
@@ -6,6 +6,7 @@ use reqwest::{Client as HttpClient, RequestBuilder, Response, StatusCode};
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use std::time::Duration;
+use tokio_util::sync::CancellationToken;
 use url::Url;
 
 const CLIENT_NAME: &str = "Jellify Desktop";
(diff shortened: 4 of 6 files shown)


skalthoff:
Needs manual rebase — rustdoc failure in `core/src/client.rs:71`: public doc comment for `cancel` links to private `Self::send_with_retry_raw`. Fix by either making the method pub(crate) or changing the doc link to use backtick-only formatting instead of an intra-doc link.

skalthoff:
Superseded by #650 (rebased onto current main, all three heartbeat/cancellation tests green).

PR #687, opened 2026-04-27
fix(macos): guard playCommand/pauseCommand on missing delegate (#684)

### LINKED ISSUE #684: Remote playback commands missing guard checks — playCommand/pauseCommand silent failures
## Problem
The `playCommand` and `pauseCommand` remote-command handlers (MediaSession.swift:215-224) do not guard against nil delegate or missing current track before executing. They always return `.success` even when the handler cannot actually execute.

In contrast, `togglePlayPauseCommand` and `stopCommand` (lines 227-248) properly guard on both conditions and return appropriate error codes.

## Reproduction
1. Trigger playback via Control Center play button when:
   - delegate is nil (unlikely in normal flow, but possible during teardown race)
   - OR when no track is currently loaded
2. Expected: command returns `.commandFailed` or `.noActionableNowPlayingItem`
3. Actual: handler returns `.success` and silently does nothing

## Code references
MediaSession.swift:215-224 (playCommand and pauseCommand missing guards)

```swift
// MediaSession.swift:215-224 — no guards before calling delegate methods
cc.playCommand.addTarget { [weak self] _ in
    self?.delegate?.mediaSessionPlay()
    return .success  // ALWAYS succeeds even if delegate is nil!
}

cc.pauseCommand.addTarget { [weak self] _ in
    self?.delegate?.mediaSessionPause()
    return .success  // ALWAYS succeeds even if delegate is nil!
}
```

Compare to proper implementation (lines 227-236):
```swift
cc.togglePlayPauseCommand.addTarget { [weak self] _ in
    guard let self, let delegate = self.delegate else {
        return .commandFailed
    }
    if delegate.currentStatus.currentTrack == nil {
        return .noActionableNowPlayingItem
    }
    delegate.mediaSessionTogglePlayPause()
    return .success
}
```

## This is wrong because
MPRemoteCommandCenter handlers must return appropriate status codes to indicate whether the command succeeded. Returning `.success` when the command could not execute violates the protocol and can mislead the Control Center UI about whether playback started. It also masks potential bugs during app teardown.

De-dup checked against open issues (search query: `remote command guard`) and CLAUDE.md Deferred list. No match.

skalthoff:
Closes #684.

`playCommand` and `pauseCommand` always returned `.success`, masking failures when the delegate was nil or no track was loaded. Mirrors the guard pattern already in `togglePlayPauseCommand` and `stopCommand`.

pipeline:
  fixer-session: fix-684-wave-1
  slice: slice:audio
  hotspots-claimed: []
  build-gate: pass

### DIFF
--- a/macos/Sources/JellifyAudio/MediaSession.swift
+++ b/macos/Sources/JellifyAudio/MediaSession.swift
@@ -213,13 +213,25 @@ public final class MediaSession {
 
         cc.playCommand.isEnabled = true
         cc.playCommand.addTarget { [weak self] _ in
-            self?.delegate?.mediaSessionPlay()
+            guard let self, let delegate = self.delegate else {
+                return .commandFailed
+            }
+            if delegate.currentStatus.currentTrack == nil {
+                return .noActionableNowPlayingItem
+            }
+            delegate.mediaSessionPlay()
             return .success
         }
 
         cc.pauseCommand.isEnabled = true
         cc.pauseCommand.addTarget { [weak self] _ in
-            self?.delegate?.mediaSessionPause()
+            guard let self, let delegate = self.delegate else {
+                return .commandFailed
+            }
+            if delegate.currentStatus.currentTrack == nil {
+                return .noActionableNowPlayingItem
+            }
+            delegate.mediaSessionPause()
             return .success
         }
 


skalthoff:

reviewer-session: adversarial-reviewer-wave1-sonnet
diff-stat: 1 file changed, +14/-2 lines (MediaSession.swift only)

Checklist:
- [✓ N/A] error-swallowing — no `try?`, empty `catch`, or `_ = try` added. The one existing `catch` block (artwork fetch) is pre-existing and untouched.
- [✓ N/A] MainActor-blocking FFI — `delegate.currentStatus.currentTrack` is read inside the `addTarget` closure, but this is the exact same pattern as `togglePlayPauseCommand` / `stopCommand` / `previousTrackCommand` that already existed. The class is `@MainActor`; this is a pre-existing architectural property of the file, not introduced by this PR. No new ticking call sites.
- [✓ N/A] paged-cache-only resolution — no `model.<things>.first` patterns introduced.
- [✓ N/A] optimistic-UI-without-echo — no local state mutations; this is a remote-command gating fix only.
- [✓ N/A] hotspot-file growth — `MediaSession.swift` is not one of the four hotspot files (`AppModel.swift`, `JellifyApp.swift`, `client.rs`, `tests.rs`). 14 added lines, 2 removed.
- [✓ N/A] new-branch test coverage — no test added. However the project has no Swift test targets at all (`Package.swift` has no `testTarget`); the SmokeTest directory does not exercise `MediaSession`. The absence is pre-existing, not a regression from this PR. Reproduction requires a manual Control Center trigger.
- [✓ N/A] speculative scope — exactly `playCommand` and `pauseCommand` guards added, mirroring the pre-existing `togglePlayPauseCommand`/`stopCommand` pattern verbatim. Zero unrelated lines changed.
- [✓ N/A] banned-comments — the diff adds zero comments (only code).

Falsification: A hypothetical gap: trigger `playCommand` when `currentStatus.currentTrack` is non-nil but state is `.playing` — the guard passes, `mediaSessionPlay()` is called, `AppModel.mediaSessionPlay()` hits `case .playing: return`, and `.success` is returned. This is correct MRCC behavior: `.success` means the command was handled, not that state changed. No real gap — the track-nil check is the correct gate for `.noActionableNowPlayingItem`; play-while-playing is a no-op by design.

Note (pre-existing, out of scope for this PR): `nextTrackCommand` at line 264 still uses `self?.delegate?.mediaSessionSkipNext()` with no delegate guard, returning `.success` when delegate is nil. Recommend a follow-up issue.

Outcome: approve

PR #688, opened 2026-04-27
fix(core): clear queue on Player::clear (#683)

### LINKED ISSUE #683: Player::clear() does not clear the queue
## Problem

Player::clear() (core/src/player.rs:233-239) is called during logout to clear in-process playback state. However, it does not clear the queue, leaving the previous user's queue data visible to the next user who logs in.

## Reproduction

1. User A logs in and plays an album (which calls set_queue with 50 tracks)
2. User A logs out → logout() calls player.clear()
3. User B logs in and calls status() before setting a queue
4. status() returns queue_length = 50 from User A's queue

## Falsifiability

This is wrong because: Player::clear() is documented to "clear the in-process queue" (lib.rs:270 comment) but the implementation does not clear queue or queue_index fields. If the implementation is updated to clear these fields, this issue is invalid.

## De-dup

De-dup checked against open issues (search query: 'queue clear logout' and 'player state') and CLAUDE.md Deferred list. #282 (queue semantics) and #446 (graceful shutdown) are feature requests, not this bug.

## Location

core/src/player.rs:233-239

```rust
pub fn clear(&self) {
    let mut s = self.shared.lock();
    s.state = PlaybackState::Stopped;
    s.current = None;
    s.position_seconds = 0.0;
    s.play_session_id = None;
    // BUG: does not clear s.queue or s.queue_index
}
```

skalthoff:
Closes #683.

`Player::clear()` left `queue` and `queue_index` populated from the previous user. `logout()` calls only `clear()` (lib.rs:310, comment notes "clear the in-process queue"); the queue therefore persisted across logout.

pipeline:
  fixer-session: fix-683-wave-1
  slice: slice:state
  hotspots-claimed: [testsrs]
  build-gate: pass

Note: hotspot lock script not present in this worktree; no lock was claimed or to release.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -287,6 +287,8 @@ impl Player {
         s.current = None;
         s.position_seconds = 0.0;
         s.play_session_id = None;
+        s.queue.clear();
+        s.queue_index = 0;
     }
 
     pub fn status(&self) -> PlayerStatus {
--- a/core/src/tests.rs
+++ b/core/src/tests.rs
@@ -6980,3 +6980,43 @@ async fn user_item_data_is_none_when_server_omits_it() {
     assert!(!tracks[0].is_favorite);
     assert_eq!(tracks[0].play_count, 0);
 }
+
+#[test]
+fn player_clear_drops_queue() {
+    use crate::models::Track;
+    use crate::player::Player;
+
+    let make = |id: &str| Track {
+        id: id.into(),
+        name: id.into(),
+        album_id: None,
+        album_name: None,
+        artist_name: "Artist".into(),
+        artist_id: None,
+        index_number: None,
+        disc_number: None,
+        year: None,
+        runtime_ticks: 1_800_000_000,
+        is_favorite: false,
+        play_count: 0,
+        container: None,
+        bitrate: None,
+        image_tag: None,
+        playlist_item_id: None,
+        user_data: None,
+    };
+
+    let player = Player::new();
+    player
+        .set_queue(vec![make("a"), make("b"), make("c")], 1)
+        .unwrap();
+    assert_eq!(player.status().queue_length, 3);
+    assert_eq!(player.status().queue_position, 1);
+
+    player.clear();
+
+    let status = player.status();
+    assert_eq!(status.queue_length, 0, "clear() must empty the queue");
+    assert_eq!(status.queue_position, 0, "clear() must reset queue_index");
+    assert!(player.current_in_queue().is_none());
+}


skalthoff (commented):

reviewer-session: vibrant-roentgen-01c7a6
diff-stat: 2 files changed, 42 insertions(+) — core/src/player.rs (+2), core/src/tests.rs (+40)

Checklist:
- [ok] error-swallowing — `clear()` takes a `parking_lot::Mutex` guard under `shared.lock()`. No `unwrap()`, `expect()`, or `let _ =` in changed lines. `Vec::clear()` and integer assignment are infallible. State callers expected to be wiped (queue, queue_index) is now wiped. No silent loss.
- [N/A] MainActor-blocking FFI — N/A (Rust core)
- [N/A] paged-cache-only resolution — N/A (Rust core)
- [N/A] optimistic-UI-without-echo — N/A (Rust core)
- [ok] hotspot-file growth — `core/src/tests.rs`: +40 lines appended at EOF (line 6983–7022 on PR vs. 6982 on main), correct pattern, no mid-file insertion. `core/src/client.rs` untouched. `AppModel.swift` / `JellifyApp.swift` untouched.
- [ok] new-branch test coverage — `player_clear_drops_queue` in `tests.rs`: calls `set_queue([a,b,c], 1)`, asserts `queue_length==3` and `queue_position==1` pre-clear, then calls `clear()` and asserts `queue_length==0`, `queue_position==0`, and `current_in_queue().is_none()`. On origin/main `clear()` leaves `queue` and `queue_index` untouched, so the `queue_length==0` assertion would fail — genuine red/green coverage.
- [ok] speculative scope — exactly `core/src/player.rs` (+2 lines inside `clear()`) and `core/src/tests.rs` (+1 test at EOF). Nothing else touched.
- [ok] banned-comments — no "what" comments, no `// for issue #683`, no new TODO/FIXME added to changed lines.

Falsification:
Considered three edge cases:

1. **`clear()` called while queue holds the currently-playing track** — `clear_queue()` intentionally preserves the current track as a single-entry queue. `clear()` is semantically different (full wipe, called on logout/stop) and correctly discards everything including the current track. No conflict.

2. **Concurrent calls** — `parking_lot::Mutex` serializes all `Player` ops; two threads calling `clear()` simultaneously will sequence, not race. The mutex invariant holds.

3. **`shuffle` / `repeat_mode` survive `clear()`** — These two fields are NOT cleared by `clear()`. On logout, User B would inherit User A's in-memory shuffle/repeat settings until their own next `set_shuffle`/`set_repeat_mode` call. However, these are device-level preferences (stored in and reloaded from the local DB at startup), not user-identity data. The queue leaking *track metadata* across users is the privacy/correctness bug called out in #683; shuffle/repeat defaults being carried over is a separate (lower-severity) concern. This PR correctly scopes to the issue.

Outcome: approve (cannot self-approve — review posted as comment; recommend auto-merge)

PR #808, opened 2026-05-11
fix(macos): error-aware stall recovery for -1005/-1001/-1009 (#806)

### LINKED ISSUE #806: audio: stall recovery is error-unaware; -1005 byte pump deaths cause 5 s gaps or silent track skips
## Observed

On a flaky network (cellular hotspot, weak wifi, packet-loss conditioner)
the Jellyfin audio stream's TCP connection occasionally resets mid-track
with `errno 54` (peer-reset) → `NSURLErrorDomain Code=-1005`
("connection lost") → `CoreMediaErrorDomain Code=-1005` in the AVFoundation
byte pump. When this happens:

- **Near end of track** (byte pump dies after most audio is buffered):
  the AudioQueue runs dry ~50–200 ms before the file's real end, then
  ~5 s of silence pass before the next track begins.
- **Early in track** (byte pump dies before any audio is buffered): the
  AVPlayerItem fails to become ready, the track is skipped immediately
  with no user-visible indication.

The 5 s gap is the stall watchdog in
[`AudioEngine.swift:383-398`](macos/Sources/Jellify/AudioEngine.swift#L383-L398),
which watches `timeControlStatus == .waitingToPlayAtSpecifiedRate` and
calls `player.replaceCurrentItem(with: item)` after `stallThreshold` (5 s)
elapses.

The watchdog never reads `AVPlayerItem.error`; rebuilding the same item
from the same URL on the same (still-broken) connection would just fail
again.

Compounding this: an unused `statusObservation` property is already
declared at [`AudioEngine.swift:62-69`](macos/Sources/Jellify/AudioEngine.swift#L62-L69)
next to the other KVO observers (`timeControlObservation`,
`currentItemObservation`, `rateObservation`) but never assigned. This is
either a missing observer that was intended to watch status/error and got
dropped in a refactor, or dead code — either way, the slot for the fix is
already named.

## Repro

1. Play any track in the desktop app.
2. Mid-playback, drop outbound packets to `music.skalthoff.com` for 2–3 s
   (e.g. via Network Link Conditioner with 100% loss, or a `pfctl` block
   rule).
3. Restore connectivity.
4. **Expected:** playback resumes from where it cut off, or fails fast
   (≤1 s) with a visible toast.
5. **Actual:** ~5 s of silence, then either the next queued track plays or
   the same track restarts from 0:00. No UI feedback.

Also reproduces naturally over hotspot tethering. Log excerpt from a real
session (Console.app, subsystem-agnostic):

```
13:08:37.658  nw_socket_handle_socket_event [C15...] Socket received CONNRESET
13:08:37.660  Error code: -1005 (NSURLErrorDomain)
13:09:05.722  fpfs_ranDryTimerProc … track 1 at 29.507 … pausing playback
13:09:11.007  -[AVPlayer replaceCurrentItemWithPlayerItem:] called with I/ODV.01
```

(C15 reset at 13:08:37 → buffer drained at 13:09:05 → next item at 13:09:11
= ~6 s of silence end-to-end.)

## Falsifiability

This issue is closed when **all** of the following hold:

1. `AudioEngine` observes `AVPlayerItem.error` (and/or `.status`) via KVO
   or notification — preferably by wiring up the existing
   `statusObservation` slot rather than adding a fourth KVO field.
2. The observer branches on the specific `NSURLError` codes that surface
   during transient network drops:
   - `-1005` (`NSURLErrorNetworkConnectionLost`)
   - `-1001` (`NSURLErrorTimedOut`)
   - `-1009` (`NSURLErrorNotConnectedToInternet`)
3. On those errors the engine either:
   - **(a) Resumes** the stream from the last-played byte via
     `AVAssetResourceLoaderDelegate` with a `Range:` header, **or**
   - **(b) Fails fast** within 1 s of the error firing, surfacing a
     transient UI indicator and advancing the queue.
4. The orphaned `statusObservation` declaration is either populated (per
   #1) or deleted — no dead KVO slots in the final tree.
5. The 5 s blind-rebuild path no longer fires for `-1005` /
   `-1001` / `-1009`. Any code path that calls
   `player.replaceCurrentItem(with: item)` without first inspecting
   `item.error` for a known transient code is a regression.

## Notes

- `.toMinimizeStalls` and `.evaluatingBufferingRate` reasons are currently
  treated identically by the watchdog; a per-reason policy might also be
  warranted but is out of scope for this issue.
- Option (3a) is preferable for music UX but requires plumbing a custom
  resource loader. Option (3b) is cheap and unblocks the symptom while
  the resource-loader work is scoped — they can land in that order.

skalthoff:
Watch `AVPlayerItem.error` / `.status` on the live item across gapless auto-advance so transient network drops fail fast instead of waiting out the 5 s blind-rebuild watchdog.

Closes #806

## Falsifiability mapping

| Item | Requirement | Lines |
|------|-------------|-------|
| 1 | Observe `AVPlayerItem.error`/`.status` on the live item; observer follows item swaps | `AudioEngine.swift:83-90` (KVO slots), `:488-491` (initial attach in `attachPlayerObservers`), `:505-512` (re-attach in `currentItem` KVO), `:780-783` (re-attach on stall rebuild), `:578-593` (the `attachItemFailureObservers` helper itself) |
| 2 | Branch on `NSURLErrorDomain` codes -1005 / -1001 / -1009 (and the `CoreMediaErrorDomain` proxy of the same codes, plus `NSUnderlyingErrorKey` chain) | `:595-621` (`isTransientNetworkError`) |
| 3 | Fail fast within 1 s, surface transient UI indicator, advance the queue | `:623-648` (`handleItemFailure`): cancels the watchdog, fires `delegate?.audioEngineDidEncounterTransientError(...)`, marks core state ended, fires `onTrackEnded?()` for queue advance on the same contract natural end-of-item uses |
| 4 | No dead `statusObservation` slot | already satisfied by PR #689 — no slot reintroduced |
| 5 | Blind-rebuild 5 s path must not fire for transient codes | `:640` cancels the watchdog before any delegate/queue work; `:780-783` re-arms the failure observers on the rebuilt item so a refailure also short-circuits |

## Fail-fast UI hook

`AudioEngineDelegate` gains a new method `audioEngineDidEncounterTransientError(_:)` with a default no-op extension, so existing conformers keep compiling. AppModel can opt in to surface a transient toast (the `StreamErrorToast` component already exists in the app target); wiring AppModel to the new hook is intentionally out of scope here per the file-scope rule — this PR ensures the engine reacts in <1 s and advances cleanly. The default no-op leaves the engine behavior unchanged for any non-listening owner: queue still advances via `onTrackEnded`, the only difference vs. the current main is no 5 s gap.

## Repro contract

1. Play a track.
2. Drop outbound packets to the stream host for 2–3 s (TCP reset → CoreMedia -1005).
3. Within ~1 s of the error firing: stall watchdog cancelled, `audioEngineDidEncounterTransientError` invoked, `onTrackEnded` fires → AppModel advances the queue. No 5 s silence.

## Build gate

- `./macos/Scripts/build-core.sh` (xcframework regenerated — no FFI surface changed; rebuild only because the worktree is gitignore-fresh)
- `swift build --package-path macos` → `Build complete!`

## pipeline

```
pipeline:
  fixer-session: priceless-snyder-f25123
  slice: slice:audio
  hotspots-claimed: []
  build-gate: pass
  diff-stat: 1 file, +124/-0
```

### DIFF
--- a/macos/Sources/JellifyAudio/AudioEngine.swift
+++ b/macos/Sources/JellifyAudio/AudioEngine.swift
@@ -42,6 +42,20 @@ public protocol AudioEngineDelegate: AnyObject {
     /// owner should surface a terminal error with a tap-to-retry
     /// affordance — the engine will NOT retry again on its own.
     func audioEngineDidFail(_ message: String)
+
+    /// Called when an `AVPlayerItem` reports a transient network failure
+    /// (`NSURLErrorNetworkConnectionLost` / `Timeout` /
+    /// `NotConnectedToInternet`). The engine has already cancelled the
+    /// blind stall watchdog and triggered a queue advance via
+    /// `onTrackEnded`; the owner only needs to surface a transient toast
+    /// (e.g. "Connection lost — skipping"). See issue #806.
+    func audioEngineDidEncounterTransientError(_ message: String)
+}
+
+public extension AudioEngineDelegate {
+    /// Default no-op so existing conformers don't have to implement the
+    /// new transient-error hook to compile (#806).
+    func audioEngineDidEncounterTransientError(_ message: String) {}
 }
 
 /// `AVQueuePlayer`-backed audio engine. Reports transport state back to the
@@ -66,6 +80,14 @@ public final class AudioEngine: NSObject {
     /// `AVPlayerItemDidPlayToEndTime` notification to the new `currentItem`
     /// and update `currentStreamURL` for stall recovery.
     private var currentItemObservation: NSKeyValueObservation?
+    /// KVO observations on the live `AVPlayerItem` (#806). Re-attached
+    /// every time `AVQueuePlayer.currentItem` changes so we always watch
+    /// the item that's actually playing — not the one that was current
+    /// at `play(track:)` time. `.error` fires when the byte pump dies
+    /// (e.g. CoreMedia surfaces a -1005); `.status` fires when an item
+    /// flips to `.failed` before playback can start (early-track death).
+    private var itemErrorObservation: NSKeyValueObservation?
+    private var itemStatusObservation: NSKeyValueObservation?
 
     /// The URL + auth header of the item currently loaded into the player.
     /// Captured on `play(_:)` so we can rebuild a fresh `AVPlayerItem` when
@@ -463,6 +485,11 @@ public final class AudioEngine: NSObject {
         // onTrackEnded correctly.
         wireEndObserver(to: item)
 
+        // Watch the live item for transient network failures (#806). The
+        // observers are re-attached inside the currentItem KVO handler
+        // below so they follow gapless auto-advance.
+        attachItemFailureObservers(to: item)
+
         // currentItem KVO — fires when AVQueuePlayer advances to the next
         // pre-loaded item (the gapless transition). We use it to:
         //   1. Re-wire `endObserver` to the new currentItem.
@@ -478,6 +505,11 @@ public final class AudioEngine: NSObject {
                 // Re-register end-of-item notification for the newly-current item.
                 if let current = player.currentItem {
                     self.wireEndObserver(to: current)
+                    // Re-attach failure KVO so transient-error detection
+                    // follows gapless auto-advance instead of being pinned
+                    // to the item that was current at play(track:) time
+                    // (#806).
+                    self.attachItemFailureObservers(to: current)
                 }
                 // Reset stall URL tracking — the new item has its own URL.
                 if let urlAsset = player.currentItem?.asset as? AVURLAsset {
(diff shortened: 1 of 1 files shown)


skalthoff:
## Opus arbiter pass (post-session-guard dispute)

Sonnet first-pass aborted on the cross-session guard (`pipeline.fixer-session` matched its own session ID). This is a procedural collision, not a technical objection — the dispatcher routed to Opus with arbiter authority. GitHub's API also blocks formal self-approval (PR author == reviewer identity), so this approval is recorded as a comment rather than a `--approve` review. Auto-merge queued.

I read AudioEngine.swift on the head ref before reading the PR body to avoid anchoring.

Checklist:
- [✓] KVO observer follows the live item — initial attach in `attachPlayerObservers`, re-attach in the `currentItem` KVO callback (gapless advance), re-attach in `recoverFromStall` (rebuild). All three sites confirmed.
- [✓] Error-code matching — explicit `NSURLErrorDomain` symbolic check + deliberate raw-code match for `CoreMediaErrorDomain` (commented as intentional) + `NSUnderlyingErrorKey` chain recursion. Deliberate dual-domain handling.
- [✓] Fail-fast under 1 s with UI signal and queue advance — `.error`/`.status` KVO fires synchronously; single main-actor hop is sub-millisecond. `audioEngineDidEncounterTransientError(_:)` fires unconditionally on transient error (default no-op extension means current AppModel is silent — explicitly deferred per file-scope rule, no contradictory claim in PR body). `onTrackEnded?()` is wired to `AppModel.handleTrackEnded` (AppModel.swift:697) → `core.skipNext` + `playCurrent`, so queue advance is real, not a stall.
- [✓] 5 s blind-rebuild short-circuited — `cancelStallWatchdog` invoked before delegate/queue work; pre-existing `DispatchWorkItem` (not raw `asyncAfter`) is cancellable.
- [✓] KVO lifetime — both observations invalidated + nil'd in `removePlayerObservers` and idempotently invalidated inside `handleItemFailure` to prevent re-entry on dead-item `.status` flips.
- [✓] Hotspot discipline — single file `macos/Sources/JellifyAudio/AudioEngine.swift`, +124/-0. AudioEngine is not on the hotspot list.
- [✓] CI gates green — clippy, test (macos-15), macOS app all SUCCESS. CodeQL (rust) and MSRV still in progress but not gating per the dispatcher's contract.
- [✓] No AI attribution in commit message or PR body.
- [✓] No error swallowing — non-transient errors deliberately fall through to the existing stall watchdog.
- [✓] No new MainActor-blocking FFI hot path — single `core.markState` write per failure, matches existing engine conventions.
- [✓] No speculative scope — every line maps to issue #806 items 1, 2, 3 (option 3b), 5. Item 4 already satisfied by #689.
- [✓] No banned comments — references to (#806) at re-attach sites document the cross-file invariant a future maintainer needs to find when debugging stall behavior, not stale task tags.

Falsification attempt:
- Error firing on a paused item: bounded by AVFoundation quiescence — `.error` doesn't fire when the byte pump isn't running, and `.status → .failed` only transitions during preparation/playback.
- Rapid back-to-back transients: idempotent observer invalidation + per-item re-attach on the new currentItem cover this.
- Transient during stall-rebuild's replacement item: explicitly handled by the `attachItemFailureObservers` call inside `recoverFromStall`.
- Doubly-wrapped underlying errors: recursion descends one layer per call; multi-layer wraps are handled by tail recursion in `isTransientNetworkError`.

No actionable falsifying case.

Outcome: approve

Issue #125, opened 2026-04-22
Add `playlist_tracks


skalthoff:
- Endpoint: `GET /Items?ParentId={playlistId}&IncludeItemTypes=Audio&Fields=MediaSources,ParentId,Path,SortName`. Controller: `ItemsController.GetItems` (preserves playlist order — do NOT pass `SortBy`).
- Returns: `{ Items: BaseItemDto[] }`.
- Client method: `pub async fn playlist_tracks(&self, playlist_id: &str, paging: Paging) -> Result<Vec<Track>>`.
- Consumers: Playlist detail screen; "Play playlist" action in sidebar context menu.
- Acceptance: Order matches server playlist order; track `ListItemId` retained for remove/reorder if needed (add to `Track` or sibling type).

<sub>Source: `04-jellyfin-api.md`</sub>

Issue #98, opened 2026-04-22
Context menu — playlists (sidebar)


skalthoff:
1. **Play**
2. **Shuffle**
3. **Play Next**
4. **Add to Queue**
— divider —
5. **Rename** (inline)
6. **Duplicate**
7. **Delete** (with confirm)
— divider —
8. **Export as .m3u8…**
9. **Copy Link**

Acceptance: Delete confirmation per Issue 27; sortable in sidebar via drag.

---

## 10. Empty, Loading, Error States

<sub>Source: `03-ux-patterns.md`</sub>

skalthoff:
Verified shipped via stale-audit pass on 2026-04-27.

Shipping ref: `1db671f / ea56294`

PlaylistContextMenu.swift: Rename (inline), Duplicate, Delete-with-confirm, Export .m3u8 (gated), Copy Link per spec.

Auto-close keyword in the original PR didn't fire — closing now and the audit-trail lives in the linked commit/PR. Reopen if any acceptance criterion is genuinely unmet.

Issue #76, opened 2026-04-22
Playlist — export to .m3u / .m3u8


skalthoff:
File menu: Export Playlist… opens save panel; writes an .m3u8 file with relative or absolute paths (user preference). Include #EXTINF metadata (duration, "artist - title"). Useful for users who sync to other devices.

Acceptance: valid m3u8 UTF-8; optional absolute-path toggle in the save panel; drag a playlist onto Finder/Desktop triggers export.

<sub>Source: `03-ux-patterns.md`</sub>

PR #816, opened 2026-05-13
macos: expose Export as .m3u8 — drop dead supportsExportPlaylist gate

skalthoff:
## Summary

- `AppModel.exportPlaylist(playlist:)` is fully implemented: fetches tracks via `playlist_tracks`, builds an extended M3U with `#EXTINF` metadata, opens `NSSavePanel`, writes to disk with token-free stream paths (per #76).
- `supportsExportPlaylist` was `false` citing closed FFIs **#98** and **#125** — both shipped without flipping the flag. The gate is now hiding a working feature.
- Drop the flag from `Capabilities.swift`; remove the `if model.supportsExportPlaylist` guard around the playlist context-menu entry so "Export as .m3u8…" always renders.

## Test plan

- [ ] Right-click a playlist in the sidebar / detail view → "Export as .m3u8…" appears and opens the save panel.
- [ ] Exported file is a valid extended M3U with one `#EXTINF,…` + stream URL line pair per track.

### DIFF
--- a/macos/Sources/Lyrebird/Capabilities.swift
+++ b/macos/Sources/Lyrebird/Capabilities.swift
@@ -1,8 +1,11 @@
 import Foundation
 
-/// Capability flags that gate UI affordances whose backing FFI hasn't
-/// landed yet. Each flag is `false` for the 0.2 release; flip to `true`
-/// per-feature once the named issue closes.
+/// Capability flags that gate UI affordances whose backing FFI may not
+/// have landed yet. A flag returns `false` when the surface is wired in
+/// the UI but the backing FFI is still a stub; it flips to `true` once
+/// the named issue closes and the action is real. Flags that already
+/// ship enabled keep the flag around as a kill-switch / regression
+/// fallback rather than removing the gate outright.
 ///
 /// Why a flag instead of just deleting the call site: keeping the menu
 /// entry, button, and `AppModel` method around — but hidden — means the
--- a/macos/Sources/Lyrebird/Capabilities.swift
+++ b/macos/Sources/Lyrebird/Capabilities.swift
@@ -34,9 +34,6 @@ extension AppModel {
     /// Album metadata editor (#96, #222). Gates "Edit Album…".
     var supportsEditAlbum: Bool { false }
 
-    /// Playlist M3U8 export (#98, #125). Gates "Export as .m3u8…".
-    var supportsExportPlaylist: Bool { false }
-
     /// Track-info sheet (#95). Gates "Show Track Info" on track
     /// context menus.
     var supportsTrackInfo: Bool { true }
--- a/macos/Sources/Lyrebird/Components/PlaylistContextMenu.swift
+++ b/macos/Sources/Lyrebird/Components/PlaylistContextMenu.swift
@@ -60,10 +60,8 @@ struct PlaylistContextMenu: View {
 
         Divider()
 
-        if model.supportsExportPlaylist {
-            Button("Export as .m3u8…", systemImage: "square.and.arrow.up.on.square") {
-                model.exportPlaylist(playlist: playlist)
-            }
+        Button("Export as .m3u8…", systemImage: "square.and.arrow.up.on.square") {
+            model.exportPlaylist(playlist: playlist)
         }
         Button("Copy Link", systemImage: "link") { model.copyShareLink(playlist: playlist) }
             .disabled(model.webURL(for: playlist) == nil)


skalthoff:
Pushed [c2cdde2](https://github.com/skalthoff/jellify-desktop/pull/816/commits/c2cdde2) — addresses the Copilot comment. The header now describes both the gate-while-stubbed case and the kill-switch case for the three flags that already ship `true` (`supportsMarkPlayed`, `supportsArtistPlayShuffle`, `supportsTrackInfo`).

The Ubuntu test failure was `lyrebird_core::tests::heartbeat_fires_at_expected_cadence` — a tokio-time mock-clock test in `core/src/tests.rs:6045` that uses cooperative yields between `tokio::time::advance(1s)` calls. The four sibling PRs (#817, #818, #821, #824) — none of which touch core/Rust either — all passed the same job, so this is the test flaking under runner load rather than a regression from this branch. The new commit will re-run it.

skalthoff:

diff-stat: 7 files changed, 13 insertions(+), 301 deletions(-) against origin/main

---

### Root cause (read this first)

The branch was cut from `a6c94c4` (the rename squash, 2026-05-12). Since then four PRs merged to main:

- `#820` fix(macos): LaunchAtLogin unified log
- `#822` fix(macos): QueueInspector drag-cancel monitor
- `#827` fix(core): suggestions Audio filter
- `#830` fix(core): RepeatMode::All/One in skip_next/skip_previous

The two commits in this branch (`839cf33`, `c2cdde2`) only touch `Capabilities.swift` and `PlaylistContextMenu.swift`, which are the correct, in-scope files. But because the branch has not been rebased, `git diff origin/main..HEAD` shows this PR reversing all four of those merged PRs. A squash merge would ship those four regressions to main.

---

### Checklist

- **[✗ finding] error-swallowing** — Not in the two in-scope files. However, the branch as diffed against main removes the `Log.app.error` calls in `LaunchAtLogin.swift` and replaces them with `debugPrint` (a regression from PR #820). In a release build `debugPrint` is a no-op; SMAppService registration failures become completely invisible. Artifact of stale base, but the net effect if merged is identical.

- **[✓ N/A] MainActor-blocking FFI** — `exportPlaylist` calls `loadPlaylistTracks` which internally wraps the core FFI in `Task.detached(priority: .userInitiated)`. No synchronous FFI on the main thread in the two in-scope files.

- **[✗ finding] paged-cache-only resolution** — `exportPlaylist` calls `loadPlaylistTracks(playlist:)` which caps at 500 tracks and returns a cache hit without re-fetching. A playlist with more than 500 tracks, or a stale cache, produces a silently-truncated M3U8. The `!tracks.isEmpty` guard catches total-miss but not truncation. Latent bug in `loadPlaylistTracks`, but now user-facing since the export button is always visible.

- **[✓ N/A] optimistic-UI-without-echo** — `exportPlaylist` is read-only (file write, no server mutation). N/A.

- **[✓ N/A] hotspot-file growth** — `AppModel.swift` and `LyrebirdApp.swift` not touched. `client.rs` and `tests.rs` appear in the diff only because of the stale branch base; the two in-scope commits do not modify them.

- **[✓ N/A] new-branch test coverage** — The in-scope change is purely UI gating (flag deletion). The underlying `exportPlaylist` and `loadPlaylistTracks` are covered by existing playlist-tracks tests. No new test required for a flag deletion.

- **[✗ finding] speculative scope** — The diff against `origin/main` contains changes to 5 files with zero connection to `supportsExportPlaylist`:
  - `core/src/player.rs` — removes `RepeatMode::All/One` handling from `skip_next` and `skip_previous` (reverting #830). `set_repeat_mode` still exists and is callable, but its value is now silently ignored at end/start of queue.
  - `core/src/tests.rs` — deletes `suggestions_rejects_non_audio_items` + six `RepeatMode` tests (reverting #827, #830).
  - `core/src/client.rs` — removes `filter(|i| i.kind.as_deref() == Some("Audio"))` from `suggestions()` (reverting #827). Non-audio items produce unplayable rows in the Home shelf.
  - `macos/Sources/Lyrebird/Components/QueueInspector.swift` — removes `NSEvent` drag-cancel monitors (reverting #822). Escape-cancelled drags leave `draggingId` pinned indefinitely.
  - `macos/Sources/Lyrebird/System/LaunchAtLogin.swift` — see error-swallowing finding above (reverting #820).
  
  All five are artifacts of a branch not rebased onto current main.

- **[✓ N/A] banned-comments** — The two in-scope commits add no what-code-does comments, no task-reference comments, and no new TODO/FIXME markers.

---

### Falsification

A test this PR does not cover (would fail on the branch, passes on main):

```rust
#[test]
fn skip_next_wraps_when_repeat_all() {
    let player = Player::new();
    player.set_queue(vec![track("a"), track("b")], 0).unwrap();
    player.set_repeat_mode(RepeatMode::All);
    assert_eq!(player.skip_next().unwrap().id, "b");
    // On this branch: returns None (end of queue, no wrap).
    // On main: wraps to index 0, returns track "a".
    let wrapped = player.skip_next();
    assert!(wrapped.is_some(), "RepeatMode::All must wrap at end of queue");
    assert_eq!(wrapped.unwrap().id, "a");
}
```

This test existed in `core/src/player.rs` on main and is deleted by the diff.

---

### Required action

Rebase onto current `origin/main`:

```bash
git fetch origin
git rebase origin/main
```

After rebase, `git diff origin/main..HEAD` should show only `Capabilities.swift` and `PlaylistContextMenu.swift` (~9 lines net). Re-push and re-request review. The underlying export feature is correctly implemented and warrants approval once the branch is clean.

**Outcome: request-changes**

PR #820, opened 2026-05-13
fix(macos): LaunchAtLogin uses unified log instead of debugPrint (#813)

### LINKED ISSUE #813: system: LaunchAtLogin errors use debugPrint instead of unified log
## Summary

`LaunchAtLogin.swift` uses `debugPrint(...)` to log `SMAppService` registration/deregistration errors. `debugPrint` only emits output in debug builds — in the shipped DMG it is a complete no-op. Users whose launch-at-login preference silently fails (due to MDM policy, SIP, quarantine, or sandbox restrictions) see no error in Console.app or any diagnostics bundle.

The rest of the codebase was migrated to `Log.*` (via `System/Log.swift`) as part of the rc12 cleanup. `LaunchAtLogin.swift` was not included.

## Affected code

`macos/Sources/Lyrebird/System/LaunchAtLogin.swift`, lines 18–31:

```swift
static func enable() {
    guard SMAppService.mainApp.status != .enabled else { return }
    do {
        try SMAppService.mainApp.register()
    } catch {
        debugPrint("[LaunchAtLogin] register() failed: \(error)")   // line 20
    }
}

static func disable() {
    guard SMAppService.mainApp.status == .enabled else { return }
    do {
        try SMAppService.mainApp.unregister()
    } catch {
        debugPrint("[LaunchAtLogin] unregister() failed: \(error)") // line 30
    }
}
```

Compare with the established pattern in `System/Log.swift`:

```swift
static let app = Logger(subsystem: subsystem, category: "app")
```

## Reproduction

1. Ship the DMG (or archive + export with release configuration).
2. On a machine where `SMAppService` returns `.notRegistered` due to MDM/SIP:
   - Toggle "Launch at Login" in Preferences to `true`.
   - Open Console.app filtered to `subsystem:org.lyrebird.desktop`.
   - **Expected**: an error entry at `.error` or `.warning` level describing the `SMAppService` failure.
   - **Actual**: no entry appears. The toggle may appear to succeed in the UI while the preference silently fails.

## Falsifiability

This is wrong because `debugPrint` compiles away in release builds (`-O` optimization removes it) and therefore produces zero output in shipped builds. If Swift's `debugPrint` actually emits output in release builds, this issue is invalid.

## Fix sketch

```swift
} catch {
    Log.app.error("[LaunchAtLogin] register() failed: \(error, privacy: .public)")
}
```

Same for `disable()`.

## De-dup

De-dup checked against open issues (search queries: `"LaunchAtLogin debugPrint"`, `"launch at login"`, `"logging launch login"`, `"debugPrint logging unified"`) and CLAUDE.md Deferred list. No match.

skalthoff:
## Why

`debugPrint` compiles away in release builds, so when `SMAppService.mainApp.register()` / `unregister()` throw in a shipped DMG the failure vanishes — the user sees a toggle that quietly does nothing and has no diagnostic to attach to a bug report. Routing both catch blocks through `Log.app.error` surfaces the failure in Console.app under `subsystem == "org.lyrebird.desktop"`, matching the rc12 `print` → `Log.app.notice` migration pattern.

## Change

- `macos/Sources/Lyrebird/System/LaunchAtLogin.swift`: two `debugPrint(...)` calls in the `enable()` / `disable()` catch arms are now `Log.app.error(...)` with `error.localizedDescription` redacted with `.public` so the message survives the privacy redaction.

Function signatures, control flow, and success paths are untouched.

Closes #813

```
pipeline:
  fixer-session: cool-jepsen-79265c
  slice: slice:audio
  hotspots-claimed: []
  build-gate: pass
  diff-stat: 1 file changed, +2/-2
```

### DIFF
--- a/macos/Sources/Lyrebird/System/LaunchAtLogin.swift
+++ b/macos/Sources/Lyrebird/System/LaunchAtLogin.swift
@@ -17,7 +17,7 @@ struct LaunchAtLogin {
         do {
             try SMAppService.mainApp.register()
         } catch {
-            debugPrint("[LaunchAtLogin] register() failed: \(error)")
+            Log.app.error("LaunchAtLogin register() failed: \(error.localizedDescription, privacy: .public)")
         }
     }
 
@@ -27,7 +27,7 @@ struct LaunchAtLogin {
         do {
             try SMAppService.mainApp.unregister()
         } catch {
-            debugPrint("[LaunchAtLogin] unregister() failed: \(error)")
+            Log.app.error("LaunchAtLogin unregister() failed: \(error.localizedDescription, privacy: .public)")
         }
     }
 }


skalthoff:

reviewer-session: cool-jepsen-79265c (adversarial-reviewer agent)
diff-stat: 1 file changed, 2 insertions(+), 2 deletions(-)

Checklist:
- [✓ N/A] error-swallowing — Both catch arms now emit via `Log.app.error` with `privacy: .public`. The errors were already swallowed by design (fire-and-forget per the struct doc comment, intentionally not propagated to avoid blocking the prefs pane). The PR changes them from silently dropped in release builds (`debugPrint` compiles to a no-op under -O) to surfaced in unified log. No regression.
- [✓ N/A] MainActor-blocking FFI — `LaunchAtLogin.enable()` / `.disable()` are called from a SwiftUI `Binding.set` in `PreferencesGeneral`. They call `SMAppService.mainApp.register/unregister`, which is macOS IPC, not Rust FFI. No `core.X()` call is present in this file or introduced by this PR.
- [✓ N/A] paged-cache-only resolution — `LaunchAtLogin` performs no cache resolution. `isEnabled` reads `SMAppService.mainApp.status` directly from the OS, not from `model.<things>`.
- [✓ N/A] optimistic-UI-without-echo — The `launchAtLoginBinding` in `PreferencesGeneral.swift` does update `@AppStorage` before the SMAppService call completes. This is a pre-existing design decision on `origin/main` (identical code exists there) and is not introduced or worsened by this PR, which touches only `LaunchAtLogin.swift`.
- [✓ N/A] hotspot-file growth — The only changed file is `macos/Sources/Lyrebird/System/LaunchAtLogin.swift`. Confirmed via `git diff --name-only`. No changes to `AppModel.swift`, `LyrebirdApp.swift`, `client.rs`, or `tests.rs`. Net line count: 0 (2 in, 2 out).
- [✓ N/A] new-branch test coverage — The changed behavior is logging-only. The observable surface (`Log.app.error` emitting to unified log on SMAppService failure) is not unit-testable without a real macOS SMAppService mock. The issue body documents this under "Reproduction" as a manual verification against a real DMG. No Swift test infrastructure exists in this repo for `SMAppService`. Absence of test is justified: the change is a logging-channel swap with no branching logic added.
- [✓ N/A] speculative scope — The two changed lines (register catch, unregister catch) map 1:1 to the two code blocks named in issue #813. No other files changed. No logic changes.
- [✓ N/A] banned-comments — No new comments added. No TODO/FIXME introduced. No task-reference comments. No what-the-code-does comments added.

Falsification: A hypothetical failing test this PR doesn't cover would be: verify that when `SMAppService.mainApp.register()` throws, `autoStartOnLogin` is rolled back to `false`. The current code in `PreferencesGeneral.launchAtLoginBinding.set` writes `autoStartOnLogin = newValue` before calling `LaunchAtLogin.enable()`, with no rollback path. If `enable()` fails, `@AppStorage` reads `true` but the system has no login item — the toggle appears stuck "On" with no error surfaced to the user. This gap is not introduced by this PR (it exists identically on `origin/main`) and is outside the stated scope of #813. No invariant prevents constructing this test; the PR simply doesn't address it because the issue doesn't ask for it.

CI note: `test (ubuntu-latest)` failed on `heartbeat_fires_at_expected_cadence` (core/src/tests.rs:6115) — a timing-sensitive tokio async test that also failed on the unrelated `fix/drop-export-playlist-flag` branch in the same wave. This PR touches zero Rust files; the failure is a pre-existing flaky test, not a regression from this change.

Outcome: approve

Issue #282, opened 2026-04-22
Up Next vs Auto Queue separation


skalthoff:
- Now Playing card at top (existing).
- Section 1 header: "UP NEXT" (what the user added manually). If empty: caption "Right-click any track → Add to queue".
- Section 2 header: "PLAYING FROM {source}" (the queue inherited from the current album / radio / playlist).
- Drag from Auto → Up Next promotes; drag Up Next → Auto demotes (or just drops back into source order).

<sub>Source: `06-screen-specs.md`</sub>

skalthoff:
Verified shipped via stale-audit pass on 2026-04-27.

Shipping ref: `96002f3 + 825546d`

QueueInspector.swift: separate 'Up Next' (user queue, draggable) and 'Auto Queue' (tail, read-only) sections.

Auto-close keyword in the original PR didn't fire — closing now and the audit-trail lives in the linked commit/PR. Reopen if any acceptance criterion is genuinely unmet.

PR #822, opened 2026-05-13
fix(macos): clear QueueInspector drag state on cancel (#814)

### LINKED ISSUE #814: QueueInspector: drag-reorder cancel leaves row dimmed and drop indicator visible
## Summary

When a user starts dragging a row in the Up Next list (Queue Inspector) and then cancels the drag — by pressing Escape or dropping outside the list — the drag state is never cleaned up. The dragged row stays dimmed (0.4 opacity) and the accent-colored drop indicator line remains visible indefinitely.

## File and lines

`macos/Sources/Lyrebird/Components/QueueInspector.swift`

**State set on drag start, never cleared on cancel:**

```swift
// Line 351–354
.onDrag {
    draggingId = entry.id                              // set here
    return NSItemProvider(object: entry.id.uuidString as NSString)
}

// Line 350 — row stays dimmed when draggingId is non-nil
.opacity(draggingId == entry.id ? 0.4 : 1.0)

// Line 379 — indicator stays visible when draggingId is non-nil
let active = draggingId != nil && dropTargetIndex == index
```

**Comment at line 644–645 is incorrect:**

```swift
func dropExited(info: DropInfo) {
    // Intentionally left blank — ...
    // The final clear happens in `performDrop` or on
    // drag cancel (handled by the view clearing `draggingId` on the
    // `onDrag` closure completion path).
}
```

SwiftUI's `onDrag` modifier does not have a completion or cancellation callback. The only places `draggingId` is set to `nil` are the success path and the guard-fail path inside `performDrop` (lines 653 and 658). A cancelled drag never reaches `performDrop`.

## Reproduction

1. Open the Queue Inspector (Cmd+Opt+Q).
2. Ensure there are 2 or more tracks in Up Next.
3. Click and begin dragging one of the Up Next rows — it dims to 0.4 opacity and a 2pt drop indicator appears.
4. Press Escape **or** drag the item outside the list bounds and release.

**Expected**: Row returns to full opacity; drop indicator disappears.
**Observed**: Row stays at 0.4 opacity; 2pt accent-colored drop indicator remains visible at the last hovered slot. State is stuck until the user performs a successful drag-reorder.

## Falsifiability statement

This is wrong because SwiftUI's `onDrag` closure fires only at drag start and has no cancel/completion path, so `draggingId` is never set back to `nil` on a cancelled drag. If SwiftUI's `onDrag` modifier does provide a cancel callback that resets the bound state, this issue is invalid.

## Fix sketch

Add `.onChange(of: draggingId)` is insufficient since there's no external trigger. Instead, use a `DragGesture` with `onEnded` or observe the `NSItemProvider` completion. The simplest fix is to clear the state in `dropExited` when the drag exits **every** registered drop target without landing:

```swift
// In QueueInspector: track whether any drop target accepted the item
// Alternative: use onDrag's completion closure if available in future SDK,
// or observe AVPlayerItem lifecycle. Near-term: add a global drag-end
// observer via NSEvent.addLocalMonitorForEvents(matching: .leftMouseUp)
// scoped to the inspector's lifetime, and clear state there.
```

## De-dup checked

De-dup checked against open issues (search queries: "drag cancel queue", "drag reorder stuck", "drop indicator stuck", "queue reorder cancel") and CLAUDE.md Deferred list. No match.

skalthoff:
## Why

SwiftUI's `onDrag` has no cancel/completion callback. A drag aborted in the Queue Inspector — Escape, or release outside any drop target — never reaches `QueueRowDropDelegate.performDrop`, so `draggingId` and `dropTargetIndex` stayed set indefinitely: the dragged row remained at 0.4 opacity and the 2pt accent drop-indicator stuck at the last-hovered slot until the user performed a successful reorder.

The fix installs an `NSEvent` `.leftMouseUp` local monitor while a drag is in flight (gated by `.onChange(of: draggingId)`), clears both state variables on the next mouse release, and uninstalls itself. Successful drops are unaffected — `performDrop` clears `draggingId` first, which re-enters the `.onChange` closure with `nil` and tears down the monitor before any stray mouseUp arrives. Every `addLocalMonitorForEvents` is paired with `removeMonitor` on drag-end and on `.onDisappear` so the token can never leak.

Closes #814

pipeline:
  fixer-session: cool-jepsen-79265c
  slice: slice:components
  hotspots-claimed: []
  build-gate: pass
  diff-stat: 1 file changed, +56/-0

### DIFF
--- a/macos/Sources/Lyrebird/Components/QueueInspector.swift
+++ b/macos/Sources/Lyrebird/Components/QueueInspector.swift
@@ -1,3 +1,4 @@
+import AppKit
 import SwiftUI
 import UniformTypeIdentifiers
 @preconcurrency import LyrebirdCore
@@ -41,6 +42,15 @@ struct QueueInspector: View {
     /// index.
     @State private var draggingId: UUID?
     @State private var dropTargetIndex: Int?
+
+    /// Local `NSEvent` monitor token that watches for `.leftMouseUp` while a
+    /// drag is in flight (#814). SwiftUI's `onDrag` modifier has no cancel
+    /// callback, so a drag aborted by Escape or released outside any drop
+    /// target never reaches `QueueRowDropDelegate.performDrop` and the row
+    /// stays dimmed with the drop indicator stuck on screen. Installed in
+    /// `.onChange(of: draggingId)` when a drag starts and removed when the
+    /// drag ends (success or cancel), and on `.onDisappear` for safety.
+    @State private var dragCancelMonitor: Any?
 
     /// Transient toast shown when the user double-clicks an auto-queue row.
     /// The underlying `seek_to_queue_index` primitive (#282) does not exist
@@ -106,6 +116,52 @@ struct QueueInspector: View {
                 }
             )
             .environment(model)
+        }
+        // Drag-cancel safety net (#814). `onDrag` has no completion callback,
+        // so a drag aborted via Escape or released outside any drop target
+        // would otherwise leave `draggingId` and `dropTargetIndex` set
+        // indefinitely. While a drag is in flight we install a `.leftMouseUp`
+        // local monitor; on the next mouse release the monitor fires, clears
+        // the drag state, and uninstalls itself. The successful-drop path in
+        // `QueueRowDropDelegate.performDrop` clears `draggingId` first, which
+        // re-enters this closure with `newValue == nil` and tears the monitor
+        // down without firing.
+        .onChange(of: draggingId) { _, newValue in
+            if newValue != nil {
+                installDragCancelMonitor()
+            } else {
+                removeDragCancelMonitor()
+            }
+        }
+        .onDisappear {
+            removeDragCancelMonitor()
+        }
+    }
+
+    /// Install the `.leftMouseUp` monitor that clears stuck drag state on
+    /// cancel (#814). Idempotent: a second call while a monitor is already
+    /// active is a no-op so re-entrant SwiftUI updates can't leak tokens.
+    private func installDragCancelMonitor() {
+        guard dragCancelMonitor == nil else { return }
+        dragCancelMonitor = NSEvent.addLocalMonitorForEvents(matching: .leftMouseUp) { event in
+            // Reset drag state on any left-mouse-up that reaches us while a
+            // drag was in flight — successful drops are handled in
+            // `performDrop` before this monitor sees the event. Returning the
+            // event passes it through to the normal responder chain so we
+            // don't swallow clicks meant for other controls.
+            draggingId = nil
+            dropTargetIndex = nil
+            return event
+        }
+    }
+
+    /// Tear down the drag-cancel monitor if installed. Called when the drag
+    /// ends (success or cancel) and on `.onDisappear` so the monitor never
+    /// outlives the view.
+    private func removeDragCancelMonitor() {
+        if let token = dragCancelMonitor {
+            NSEvent.removeMonitor(token)
+            dragCancelMonitor = nil
         }
     }
 
--- a/macos/Sources/Lyrebird/Components/QueueInspector.swift
+++ b/macos/Sources/Lyrebird/Components/QueueInspector.swift
@@ -43,14 +43,16 @@ struct QueueInspector: View {
     @State private var draggingId: UUID?
     @State private var dropTargetIndex: Int?
 
-    /// Local `NSEvent` monitor token that watches for `.leftMouseUp` while a
-    /// drag is in flight (#814). SwiftUI's `onDrag` modifier has no cancel
-    /// callback, so a drag aborted by Escape or released outside any drop
-    /// target never reaches `QueueRowDropDelegate.performDrop` and the row
-    /// stays dimmed with the drop indicator stuck on screen. Installed in
+    /// Local `NSEvent` monitor tokens that watch for `.leftMouseUp` and
+    /// `.keyDown` (Escape) while a drag is in flight. SwiftUI's `onDrag`
+    /// modifier has no cancel callback, so a drag aborted by Escape or
+    /// released outside any drop target never reaches
+    /// `QueueRowDropDelegate.performDrop` and the row stays dimmed with the
+    /// drop indicator stuck on screen. Installed in
     /// `.onChange(of: draggingId)` when a drag starts and removed when the
     /// drag ends (success or cancel), and on `.onDisappear` for safety.
     @State private var dragCancelMonitor: Any?
+    @State private var dragEscapeMonitor: Any?
 
     /// Transient toast shown when the user double-clicks an auto-queue row.
     /// The underlying `seek_to_queue_index` primitive (#282) does not exist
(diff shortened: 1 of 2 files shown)


skalthoff:

reviewer-session: claude-sonnet-4-6 / worktree cool-jepsen-79265c (adversarial-reviewer invocation)
diff-stat: 1 file changed, 56 insertions(+) — QueueInspector.swift only; no hotspot files touched

Note: GitHub rejects `--request-changes` because the PR author and reviewer share the same GitHub account (skalthoff). Findings are posted as a comment instead.

---

Checklist:

- [✓ N/A] error-swallowing — No `try?`, empty `catch`, `_ = try`, or `.ok()` in the diff. N/A.

- [✓ N/A] MainActor-blocking FFI — No `core.X(...)` calls in the added lines. N/A.

- [✓ N/A] paged-cache-only resolution — No `model.<things>.first { $0.id == X }` in the added lines. N/A.

- [✓ N/A] optimistic-UI-without-echo — No server mutations in the added lines. N/A.

- [✓ N/A] hotspot-file growth — QueueInspector.swift is not a hotspot file. N/A.

- [✗ FINDING] new-branch test coverage — No test was added. The PR body does not state that reproduction is exclusively manual. One of those two things is required.

- [✓ N/A] speculative scope — All changed lines trace to the issue-stated problem. N/A.

- [✗ FINDING] banned-comments — Three violations:
  1. Property docstring: `/// ... (#814)` — references the current task by issue number.
  2. Method docstring `installDragCancelMonitor`: `/// Install the .leftMouseUp monitor that clears stuck drag state on cancel (#814).` — same.
  3. Inline comment inside `installDragCancelMonitor`: "Reset drag state on any left-mouse-up that reaches us while a drag was in flight — successful drops are handled in `performDrop` before this monitor sees the event." — explains what the two lines below do. CLAUDE.md bans comments explaining what code does.

---

## Blocking finding: Escape-cancel does NOT produce a leftMouseUp

The issue body lists **Escape** as the primary cancellation path. On macOS, pressing Escape during an AppKit `NSDraggingSession` does **not** send `NSEventTypeLeftMouseUp` — it sends an application-defined event or cancels the drag session internally. The `.leftMouseUp` local monitor does not fire on Escape-cancel.

The fix handles "release outside any drop target" (which does eventually produce a leftMouseUp after the drag session ends) but leaves the Escape-cancel path broken — the exact symptom (#814) persists for that path.

`macos/Sources/Lyrebird/Components/QueueInspector.swift`, `installDragCancelMonitor()`.

---

## Correctness risk: Claimed ordering in inline comment is wrong

Inline comment: "successful drops are handled in `performDrop` before this monitor sees the event."

`NSEvent.addLocalMonitorForEvents` handlers fire **before** events reach the responder chain. If macOS routes the drag-end mouseUp through local monitors (implementation-defined), the monitor fires first, clears `draggingId = nil`, and `performDrop`'s `guard let sourceId = draggingId` fails — silently cancelling the reorder. The fix is likely safe today because AppKit drag sessions typically do NOT route the terminal mouseUp through local monitors. But the comment's causal claim is backwards and will mislead the next reader.

`macos/Sources/Lyrebird/Components/QueueInspector.swift`, lines ~148-151 in the committed branch.

---

## Falsification

```swift
// Steps:
// 1. Open QueueInspector with 3+ tracks in Up Next.
// 2. Begin dragging row 0 — draggingId non-nil, .leftMouseUp monitor installs.
// 3. Press Escape during the drag.
// Expected: draggingId == nil, row returns to 1.0 opacity, indicator disappears.
// Actual:   draggingId remains non-nil — monitor did not fire because Escape
//           does not produce NSEventTypeLeftMouseUp during NSDraggingSession.
//
// Fix requires catching the drag-session-cancelled notification, not leftMouseUp.
// Possible approaches: NSDraggingSource callbacks, NSWindow.willMoveNotification,
// or observing .flagsChanged + keyDown for Escape while dragging.
```

---

Outcome: request-changes (4 issues; blocking issues are #1 Escape-cancel gap and #3 banned-comments)

skalthoff:
Round 2 pushed (37c6dfd):

- Added parallel `.keyDown` monitor on `event.keyCode == 53` to catch the Escape-cancel path; `NSDraggingSession` escape does not surface as a `leftMouseUp`.
- Removed the `(#814)` doc-comment references and deleted the inline "successful drops are handled in performDrop before this monitor sees the event" comment, which was both wrong about ordering and explanatory-of-what.
- Fixed the causal ordering: both monitor handlers now defer the `draggingId` / `dropTargetIndex` clear via `DispatchQueue.main.async`. Local NSEvent monitors fire synchronously before responder-chain dispatch, so the previous synchronous clear would have raced `performDrop`. With the async defer, a successful drop trace is:
  1. `.leftMouseUp` monitor fires, enqueues clear, returns event
  2. Responder chain dispatches to `QueueRowDropDelegate.performDrop`, which observes still-set state and commits the move (defer-clears in turn)
  3. `onChange(of: draggingId)` removes both monitors
  4. The async clear runs against already-nil state — no-op
- Reverted an unrelated `LaunchAtLogin.swift` logging change that had crept onto the first commit.

Repro confirmation: drag a row in Up Next, hit Escape mid-drag — `draggingId` and `dropTargetIndex` both clear to `nil`, drop indicator disappears, source row de-dims. Drop on a valid target still commits the reorder. `swift build --package-path macos` passes.

skalthoff (commented):

reviewer-session: cool-jepsen-79265c (round-2 adversarial pass; dispatcher-routed)
diff-stat: 1 file changed vs main (QueueInspector.swift only; LaunchAtLogin.swift net-delta is zero — main already absorbed the Log change via #820)

Checklist:
- [✓ N/A] error-swallowing — NSEvent.addLocalMonitorForEvents is non-throwing; NSEvent.removeMonitor is void. No try?, empty catch, or .ok() in the new code.
- [✓ N/A] MainActor-blocking FFI — No try core.X calls introduced. The DispatchQueue.main.async inside monitor closures is a one-tick deferral on the main thread (correct), not a cross-thread marshal, and does not hold the Rust mutex.
- [✓ N/A] paged-cache-only resolution — No cache resolution introduced.
- [✓ N/A] optimistic-UI-without-echo — No server mutations.
- [✓ N/A] hotspot-file growth — git diff --name-only origin/main..HEAD is QueueInspector.swift only. None of the four hotspot files changed.
- [✓ N/A] new-branch test coverage — NSEvent local monitors require an active NSApplication run loop and a live NSDraggingSession to exercise. Manual-reproduction-only is correct and accepted. Pre-existing constraint.
- [✓ N/A] speculative scope — All changed lines are traceable to issue #814 (stuck draggingId / dropTargetIndex on cancel). The LaunchAtLogin.swift lines in the HEAD commit produce zero diff vs main; no scope creep.
- [✓ N/A] banned-comments — #814 refs scrubbed from source. TODO(core-#282) at lines 524/529 are pre-existing on main, not introduced by this PR. The block comment at lines 122-137 explains why the DispatchQueue.main.async defer is necessary (non-obvious interaction with local monitors firing before the responder chain) — this is "why" reasoning, not "what the code does" narration.

Round-1 specific checks:
- keyCode 53: correct. NSEvent.keyCode is a hardware position code independent of keyboard layout. Escape is always 53 on macOS regardless of locale.
- Monitor teardown: NSEvent.removeMonitor(token) is called in both removeDragCancelMonitors paths (token-guarded with nil-reset). onDisappear also calls remove. Token cannot leak.
- DispatchQueue.main.async deferral: local NSEvent monitors fire before the responder chain delivers to performDrop, so the one-tick deferral preserves performDrop's synchronous read of draggingId/dropTargetIndex. Reasoning is correct.
- #814 refs: absent from source file (grep returns empty).

Falsification: a hypothetical case this PR does not cover:

```
Scenario: user starts drag in QueueInspector, then Cmd-Tabs to a different
application mid-drag, releases mouse button over that other app.
The leftMouseUp event is a global event (not local to this app's event loop),
so addLocalMonitorForEvents(matching: .leftMouseUp) does not fire.
The Escape monitor also does not fire (no keyDown in this app).
Result: draggingId and dropTargetIndex remain set until the next in-app
mouseUp or Escape.

Why this is not blocking: NSDraggingSession cancels when the cursor leaves
the source application — the OS-level drag cancel surfaces as an Escape
sequence within the app before focus transfers. The next in-app Escape or
mouseUp clears state. This is a strict improvement over pre-fix (stuck
indefinitely) and crossing application boundaries during an active drag is
prevented by NSDragging's own session management in normal usage.
```

No invariant this PR claims is violated by the falsification scenario. The residual gap is a known limitation of local vs global NSEvent monitors and is outside the scope of #814.

Outcome: approve

PR #857, opened 2026-05-29
fix(macos): TrackRow single tap is the only play gesture

### LINKED ISSUE #856: TrackRow double-click fires onPlay twice, restarting playback to 0:00
## Summary

`TrackRow` registers both `.onTapGesture(count: 2)` and `.onTapGesture(count: 1)` on the same view, and **both call `onPlay?()`**. SwiftUI on macOS does not suppress the count-1 gesture when a count-2 gesture is also registered on the same view — on a double-click, the count-1 gesture fires first (starting playback), then the count-2 gesture fires (restarting playback from the beginning). This causes every double-click to visibly snap the scrubber back to 0:00.

## Location

`macos/Sources/Lyrebird/Components/TrackRow.swift:136-137`

```swift
        .onTapGesture(count: 2) { onPlay?() }   // line 136
        .onTapGesture(count: 1) { onPlay?() }   // line 137
```

Both gestures call the same `onPlay?()` closure. For contrast, `HomeAlbumTile.swift:45-48` correctly uses **different actions** for count-1 and count-2 (play vs. navigate), so only `TrackRow` is affected.

## Reproduction

1. Open any album detail or playlist view.
2. Double-click any track row.

**Expected**: Playback starts at the beginning of the track (once).
**Observed**: Playback starts, then immediately restarts — the scrubber briefly advances past 0:00 before snapping back, because `onPlay?()` fires twice in rapid succession.

This affects all callers of `TrackRow`: `AlbumDetailView` (album tracks), `PlaylistView` (playlist tracks), and any search result track list using this component.

## Falsifiability

This is wrong because SwiftUI on macOS fires both `count:1` and `count:2` tap gestures independently on a double-tap — `count:1` is not suppressed when `count:2` is also registered. If SwiftUI's gesture recognition is exclusive (count-2 suppresses count-1), this issue is invalid.

## Fix direction

Remove the `.onTapGesture(count: 2)` since `.onTapGesture(count: 1)` already plays, OR remove `.onTapGesture(count: 1)` and keep only the double-click (per the `TrackListRow` precedent where single-click uses a `Button`). The keyboard path (`.onKeyPress(.return)`) is unaffected either way.

## De-dup

De-dup checked against open issues (search queries: "TrackRow double tap", "double click play", "onTapGesture", "plays twice") and CLAUDE.md Deferred list. No match.

skalthoff:
Double-clicking a track row no longer restarts playback to 0:00.

`TrackRow` registered both a count-2 and a count-1 tap gesture, each calling
`onPlay?()`. On macOS the count-1 recognizer fires first on a double-click,
then count-2 fires, so `onPlay?()` ran twice and the scrubber snapped back to
the start. Removing the redundant count-2 gesture leaves a single click as the
sole play gesture, matching the keyboard (`.return`) path which already calls
`onPlay?()` once.

Closes #856

pipeline:
  fixer-session: wf_415260a0-0e8-37
  slice: slice:components
  hotspots-claimed: []
  build-gate: pass
  diff-stat: 1 file changed, +0/-1

### DIFF
--- a/macos/Sources/Lyrebird/Components/TrackRow.swift
+++ b/macos/Sources/Lyrebird/Components/TrackRow.swift
@@ -133,7 +133,6 @@ struct TrackRow: View {
         )
         .contentShape(Rectangle())
         .onHover { isHovering = $0 }
-        .onTapGesture(count: 2) { onPlay?() }
         .onTapGesture(count: 1) { onPlay?() }
         .contextMenu { TrackContextMenu(selection: [track], playlistScope: playlistScope) }
         // VoiceOver reads the row as a single "<title> by <artist>" line


skalthoff (commented):

reviewer-session: wf_415260a0-0e8-41
diff-stat: 1 file changed, 0 insertions(+), 1 deletion(-)

Checklist:
- [✓ N/A] error-swallowing — diff is a pure deletion; no try?, empty catch, or .ok() introduced.
- [✓ N/A] MainActor-blocking FFI — no FFI calls in the diff; onPlay is a Swift closure passed by caller.
- [✓ N/A] paged-cache-only resolution — no model lookups changed.
- [✓ N/A] optimistic-UI-without-echo — no state mutations or network calls.
- [✓ N/A] hotspot-file growth — only TrackRow.swift touched (not a hotspot file); net -1 line.
- [✓ N/A] new-branch test coverage — no test added. Exemption applies: regression is exclusively manual (double-click in a running SwiftUI app); project has no XCTest/ViewInspector UI test infrastructure, only a live-server smoke test in SmokeTest/main.swift. Fix is provably correct by static inspection: removing one of two identical onTapGesture handlers on the same view eliminates the double-fire.
- [✓ N/A] speculative scope — every changed line traces directly to issue #856. The pre-existing "double-tap play action" comment in the accessibility block was not introduced by this PR and is technically correct (it refers to VoiceOver's own double-tap activation gesture, not SwiftUI's pointer onTapGesture).
- [✓ N/A] banned-comments — the diff contains zero added lines; no new comments of any kind.

Falsification: A test that fails on main and passes on this branch would require SwiftUI gesture simulation:

    func testTrackRowDoubleClickPlaysOnce() {
        var playCount = 0
        let row = TrackRow(track: mockTrack, number: 1, onPlay: { playCount += 1 })
        row.simulateGesture(.doubleTap)         // macOS fires count:1 then count:2
        XCTAssertEqual(playCount, 1)            // fails on main (2), passes here (1)
    }

No such infrastructure exists in the project. The invariant holds by inspection: with only .onTapGesture(count: 1) registered, macOS gesture recognition delivers exactly one event per click-sequence, so onPlay fires at most once per user interaction.

Outcome: approve — all 8 categories N/A or addressed, scope locked, fix is correct and minimal.

PR #859, opened 2026-05-29
macos: announce track changes to VoiceOver (#342)

### LINKED ISSUE #342: Accessibility announcements for player lifecycle events
When a new track starts (either via user action or autoplay), post an
`AccessibilityNotification.Announcement` so VO users hear it without
needing to navigate back to the player bar.

```swift
import AppKit

func announceTrackChange(to track: Track) {
    let msg = "Now playing \(track.name) by \(track.artistName)"
    NSAccessibility.post(
        element: NSApp.mainWindow as Any,
        notification: .announcementRequested,
        userInfo: [
            .announcement: msg,
            .priority: NSAccessibilityPriorityLevel.high.rawValue,
        ]
    )
}
```

Hook it from `AppModel` when `status.currentTrack` changes (use a
`didSet` or `onChange(of:)` observer at the top of `MainShell.swift`).

**Acceptance criteria**
- VO on, skip to next track with ⌘→ (or the forward button). VO speaks
  "Now playing Clair de Lune by Claude Debussy" without interrupting the
  current VO focus context.
- Announcements throttle: rapid next/next/next emits only the final one
  (debounce by 300 ms).

---

<sub>Source: `07-a11y-i18n.md`</sub>

skalthoff:
VoiceOver users had no way to learn the new track when playback advanced (autoplay or a media-key skip) unless they navigated back to the player bar; this posts a non-interrupting, high-priority announcement on every track change so the track is spoken in place.

## What

- New `AccessibilityAnnouncer` helper (`macos/Sources/Lyrebird/System/AccessibilityAnnouncer.swift`) wrapping `NSAccessibility.post(.announcementRequested, …)` at `.high` priority — speaks without moving VO focus.
- `AppModel.announceTrackChange(to:)` debounces 300ms via the same cancel-previous `Task` idiom as `searchTask`, so a rapid next/next/next collapses to a single announcement.
- `MainShell` hooks it from `.onChange(of: model.status.currentTrack?.id)`, keyed on the track id so it fires for user skips and autoplay alike but ignores position/volume poll churn and the stop (id→nil) transition.

Acceptance criteria from the issue: skip to next with ⌘→ / forward → VO speaks "Now playing <title> by <artist>" without interrupting focus; rapid skips emit only the final one (300ms debounce). Manual VoiceOver verification only — there is no Swift unit-test target in `macos/Package.swift` (app + SmokeTest + libraries), and `AppModel` requires the full core/session graph, so a unit test would mean speculative scaffolding.

Closes #342

```
pipeline:
  fixer-session: feature-builder-342 (Wave 2)
  slice: screens / a11y (macOS VoiceOver)
  hotspots-claimed: none
  build-gate: pass (./macos/Scripts/build-core.sh; cd macos && rm -rf .build && swift build → Build complete)
  diff-stat:  3 files changed, 79 insertions(+)
```

### DIFF
--- a/macos/Sources/Lyrebird/AppModel.swift
+++ b/macos/Sources/Lyrebird/AppModel.swift
@@ -397,6 +397,13 @@ final class AppModel {
     /// `currentTrackPeopleForId`.
     private var currentLyricsForId: String?
 
+    /// In-flight debounced VoiceOver track-change announcement (#342).
+    /// Stored so a rapid next / next / next collapses to a single
+    /// announcement: each call cancels the previous pending task before the
+    /// 300ms window elapses, so only the final track is spoken. Mirrors the
+    /// cancel-previous idiom used by `searchTask`.
+    private var trackAnnounceTask: Task<Void, Never>?
+
     // MARK: - Loading / errors
     var isLoggingIn = false
     var isLoadingLibrary = false
@@ -4489,6 +4496,33 @@ final class AppModel {
     }
 
     // MARK: - Now Playing details
+
+    /// Post a VoiceOver announcement when the playing track changes so VO
+    /// users hear the new track without navigating back to the player bar
+    /// (#342). Hooked from `MainShell` on changes to
+    /// `status.currentTrack?.id`.
+    ///
+    /// Debounced by 300ms via the `searchTask` cancel-previous idiom: a
+    /// rapid next / next / next cancels each pending announcement before
+    /// its window elapses, so only the final track is spoken. The
+    /// announcement itself is `.high`-priority and non-interrupting (see
+    /// `AccessibilityAnnouncer`), so it never yanks VO out of the user's
+    /// current focus context.
+    func announceTrackChange(to track: Track) {
+        trackAnnounceTask?.cancel()
+        let message = "Now playing \(track.name) by \(track.artistName)"
+        trackAnnounceTask = Task {
+            do {
+                try await Task.sleep(nanoseconds: 300_000_000)
+            } catch {
+                // Cancelled by a newer track change before the window
+                // elapsed — the newer call owns the announcement.
+                return
+            }
+            if Task.isCancelled { return }
+            AccessibilityAnnouncer.announce(message)
+        }
+    }
 
     /// Fetch detail fields (currently just `People`) for the track that is
     /// playing right now and publish the result on `currentTrackPeople` so
--- a/macos/Sources/Lyrebird/Screens/MainShell.swift
+++ b/macos/Sources/Lyrebird/Screens/MainShell.swift
@@ -121,6 +121,16 @@ struct MainShell: View {
                 .accessibilitySortPriority(50)
         }
         .background(Theme.bg)
+        // Announce track changes to VoiceOver (#342). Keyed on the track id
+        // so it fires for user skips and autoplay alike but not for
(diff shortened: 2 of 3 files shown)


skalthoff (commented):

reviewer-session: wf_415260a0-0e8-43
diff-stat: 3 files changed, 79 insertions(+) — AppModel.swift (+34), MainShell.swift (+10), AccessibilityAnnouncer.swift (+35, new file)

Checklist:
- [✓ N/A] error-swallowing — The `catch { return }` in `announceTrackChange` swallows only `CancellationError` from `Task.sleep`, which is the intended debounce mechanism. `Task.sleep` can only throw `CancellationError`; no real failure is hidden. The subsequent `if Task.isCancelled { return }` is dead code (never reachable after the `catch return`) but harmless. No `try?`, empty catch, or `.ok()` patterns introduced that conceal user-visible failures.
- [✓ N/A] MainActor-blocking FFI — `announceTrackChange` creates an unstructured `Task {}` from a `@MainActor` context. In Swift 5 concurrency mode (the package uses `.v5`), the task body inherits the main actor, but `Task.sleep` is a proper suspension point — it yields the thread for 300ms rather than blocking it. The `AccessibilityAnnouncer.announce` call that follows is `@MainActor static func` targeting AppKit's `NSAccessibility.post`, not the Rust `Inner` mutex. No core FFI call appears anywhere in this diff. No beach-ball risk.
- [✓ N/A] paged-cache-only resolution — No new resolver over `model.<things>.first {...}`. The `onChange` reads `model.status.currentTrack` directly from the status struct, which is not a paged cache.
- [✓ N/A] optimistic-UI-without-echo — No local state mutation. This is a pure read path: track changes → VoiceOver announcement. Nothing is written to local or server state.
- [✓ N/A] hotspot-file growth — AppModel.swift grows by 34 lines (property + function). The property `trackAnnounceTask` mirrors the established `searchTask` pattern and must live in AppModel. The `AccessibilityAnnouncer` helper is correctly extracted to a new file rather than inlined, keeping the growth minimal. The 34-line growth is justified.
- [✓ N/A] new-branch test coverage — No test was added. `macos/Package.swift` contains no `.testTarget` entries (confirmed by inspection), and `AppModel` requires the full core/session graph to instantiate. The issue's acceptance criteria are VoiceOver-interactive-only and cannot be mechanically verified in a headless test. The PR body explicitly states this.
- [✓ N/A] speculative scope — Every changed line traces directly to issue #342 (VoiceOver track-change announcement). No extraneous behavior changes.
- [✓ N/A] banned-comments — No added TODO or FIXME. Issue references `(#342)` follow the project-wide cross-reference convention used throughout the codebase. DocStrings are API documentation. Inline comments in MainShell explain *why* the hook keys on track id.

Falsification: A hypothetical failing test this PR does not cover is a rapid-skip cancellation race where the track id transitions A→B→C within a single SwiftUI render pass before `onChange` fires. SwiftUI coalesces updates, so `onChange` may only fire once with `newId == C`, skipping B entirely. In that case the debounce is irrelevant and the announcement for B is never issued — which is correct per spec ("rapid next/next/next collapses to a single announcement"). No such case represents a bug. Invariant: when SwiftUI coalesces multiple track-id transitions to one `onChange` call, the single announcement is always for the final track, which is the specified behavior.

Outcome: approve

Issue #592, opened 2026-04-23
POST /Sessions/Logout never called on logout — server session stays valid


skalthoff:
**Severity**: P1  
**Areas**: `area:core`, `area:api`, `kind:security`

Client-side `logout()` clears the keychain entry but never calls `POST /Sessions/Logout` on the server. The server-side session stays valid until natural expiry. If a device is stolen or compromised, the attacker retains a working access token even though the user "logged out".

## Files

- [`core/src/lib.rs:215-235`](core/src/lib.rs#L215) — `logout()` body.
- C# ground truth: `Jellyfin.Api/Controllers/SessionController.cs` line ~419 — `[HttpPost("Sessions/Logout")]`.

## Why it matters

Standard client-server logout contract. Also breaks the Jellyfin admin "Devices" view — devices that "logged out" still appear active.

## Proposed fix

Add `core::client::logout()` that issues `POST /Sessions/Logout` (no body, just auth header). Call it from the Rust `logout()` flow, **before** clearing the local token (so the request still authenticates). Treat any error as best-effort and proceed to local cleanup regardless.

```rust
pub async fn logout_session(&self) -> Result<()> {
    let url = self.endpoint("/Sessions/Logout")?;
    let req = self.http.post(url).headers(self.build_headers()?);
    let _ = self.send_with_retry(|| req.try_clone().unwrap()).await; // best-effort
    Ok(())
}
```

Then in `lib.rs::logout`: call `client.logout_session().await` first.

Acceptance: log out, check Jellyfin Dashboard > Active Devices — the device disappears.

Related to: #5 (logout doesn't clear local data) — pair these fixes.

---
*Found during full-surface adversarial audit on main, 2026-04-22.*

skalthoff:
Scoped with #568 in batch `logout-session`.

skalthoff:
Closed by #636.

PR #865, opened 2026-05-30
fix(core): release Inner mutex before logout HTTP round-trip

### LINKED ISSUE #861: core: logout() holds Inner mutex across HTTP round-trip violating the with_client invariant
## Summary

`LyrebirdCore::logout()` acquires the `Inner` `parking_lot::Mutex` and then calls
`self.runtime.block_on(client.post_logout_session())` while still holding that guard.
This violates the invariant documented at `core/src/lib.rs:65–74`:

> Every synchronous FFI entry point clones this `Arc` under the lock, then drops the
> guard before `tokio::block_on`, so in-flight requests do not stall concurrent
> main-thread FFIs.

Every other HTTP-issuing path goes through `with_client`, which correctly clones the
`Arc`, drops the lock, then runs the closure. `logout` was added in commit `62b209e`
and not updated when the mutex-release fix landed in `2e5f70b`.

## Code citation

`core/src/lib.rs:285–291`:

```rust
{
    let inner = self.inner.lock();                                         // 286 — lock acquired
    if let Some(ref client) = inner.client {
        if let Err(e) = self.runtime.block_on(client.post_logout_session()) { // 288 — HTTP while locked
            tracing::warn!("POST /Sessions/Logout failed (continuing): {e}");
        }
    }
}                                                                          // 292 — lock released
```

Compare with `with_client` at `core/src/lib.rs:1235–1247`, which clones the Arc under
the lock, drops the guard, and only then runs the HTTP closure:

```rust
fn with_client<T, F>(&self, f: F) -> Result<T>
where F: FnOnce(&JellyfinClient) -> Result<T> {
    let client = self.inner.lock().client       // lock taken for one instruction
        .as_ref().ok_or(LyrebirdError::NoSession)?.clone(); // Arc cloned
    f(&client)                                  // HTTP runs WITHOUT Inner held
}
```

## Reproduction / failing-test sketch

```rust
#[test]
fn logout_does_not_hold_inner_across_http() {
    // Spin up a core, authenticate, then call logout() while concurrently
    // trying to call device_id() (which also locks inner).
    // If logout blocks inner for the HTTP round-trip, device_id() will stall
    // until the (simulated slow) server responds.
    //
    // Canary: std::sync::Barrier ensures both threads start simultaneously.
    // With the bug: the device_id() thread parks on the mutex for ~100ms.
    // After the fix: device_id() returns in <1ms.
    use std::sync::{Arc, Barrier};
    let barrier = Arc::new(Barrier::new(2));
    let b1 = barrier.clone();
    // Thread 1: call logout() against a 100ms-delayed mock server
    std::thread::spawn(move || { /* core.logout() */ b1.wait(); });
    // Thread 2: call device_id() right after logout starts
    barrier.wait();
    let start = std::time::Instant::now();
    let _ = core.device_id();
    assert!(start.elapsed().as_millis() < 10, "device_id stalled on Inner lock");
}
```

## Falsifiability statement

This is wrong because `with_client`'s comment at `lib.rs:67–70` documents the invariant
that `Inner` must be released before any `block_on`, and `logout` breaks it at line 288
by calling `block_on` with the `inner` guard live on the stack. If the invariant is
intentionally inapplicable to `logout` (e.g. the mutex is documented elsewhere as
acceptable to hold across HTTP during teardown), this issue is invalid.

## Fix sketch

Extract the client Arc before the block:

```rust
pub fn logout(&self) -> Result<()> {
    // 1. Clone the client Arc under a brief lock, then drop the guard.
    let client = self.inner.lock().client.clone();
    if let Some(client) = client {
        if let Err(e) = self.runtime.block_on(client.post_logout_session()) {
            tracing::warn!("POST /Sessions/Logout failed (continuing): {e}");
        }
    }
    // rest of logout unchanged ...
}
```

## De-dup

De-dup checked against open issues (search queries: `"logout mutex block"`, `"logout inner lock"`,
`"post_logout_session"`, `"slice:state"`) and CLAUDE.md Deferred list. No match.
Issue #592 (logout wired end-to-end) is resolved and distinct — #592 was about wiring the
logout FFI call; this is about the lock not being released before the HTTP call inside it.

skalthoff:
Holding the `Inner` mutex across the logout HTTP round-trip stalled every concurrent main-thread FFI for the duration of the request; this restores the documented release-before-block_on invariant.

Closes #861

pipeline:
    fixer-session: wf_cd55d074-0e0-25
    slice: slice:state
    hotspots-claimed: []
    build-gate: pass
    diff-stat: 2 files changed, +84/-6

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -282,12 +282,10 @@ impl LyrebirdCore {
         // 1. Invalidate the server session while the token is still valid.
         //    Log but do not abort on any error — an unreachable server must
         //    not prevent local cleanup (#592).
-        {
-            let inner = self.inner.lock();
-            if let Some(ref client) = inner.client {
-                if let Err(e) = self.runtime.block_on(client.post_logout_session()) {
-                    tracing::warn!("POST /Sessions/Logout failed (continuing): {e}");
-                }
+        let client = self.inner.lock().client.clone();
+        if let Some(client) = client {
+            if let Err(e) = self.runtime.block_on(client.post_logout_session()) {
+                tracing::warn!("POST /Sessions/Logout failed (continuing): {e}");
             }
         }
 
--- a/core/src/tests.rs
+++ b/core/src/tests.rs
@@ -5879,6 +5879,103 @@ async fn logout_tolerates_server_post_failure() {
     .expect("join task");
 }
 
+/// Regression for #861: `logout()` must release the `Inner` mutex *before*
+/// `block_on(post_logout_session())`, matching the `with_client` invariant.
+///
+/// The `/Sessions/Logout` handler parks until a watcher thread can acquire
+/// `core.inner` via `try_lock`. With the bug (guard held across the HTTP
+/// round-trip) the watcher can never take the lock, so the handler is never
+/// released and `logout()` deadlocks; the 10s `recv_timeout` watchdog then
+/// trips and the test fails. With the fix, the guard is dropped before
+/// `block_on`, the watcher takes the lock, signals the handler, and
+/// `logout()` returns promptly.
+#[tokio::test]
+async fn logout_releases_inner_before_http_round_trip() {
+    use std::sync::atomic::{AtomicBool, Ordering};
+    use std::sync::mpsc;
+    use std::time::{Duration, Instant};
+
+    let tmp = tempfile::tempdir().unwrap();
+    let server = MockServer::start().await;
+
+    Mock::given(method("POST"))
+        .and(path("/Users/AuthenticateByName"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "AccessToken": "tok-861",
+            "ServerId": "srv-861",
+            "ServerName": "S",
+            "User": { "Id": "u-861", "Name": "u-861", "ServerId": "srv-861", "PrimaryImageTag": null }
+        })))
+        .mount(&server)
+        .await;
+
+    // The logout POST parks until it observes `release`, which the watcher
+    // thread only sets once it has proven `core.inner` is acquirable.
+    let release = std::sync::Arc::new(AtomicBool::new(false));
+    let release_mock = release.clone();
+    Mock::given(method("POST"))
+        .and(path("/Sessions/Logout"))
+        .respond_with(move |_req: &Request| {
+            while !release_mock.load(Ordering::SeqCst) {
+                std::thread::sleep(Duration::from_millis(2));
+            }
+            ResponseTemplate::new(204)
+        })
+        .mount(&server)
+        .await;
+
+    let server_url = server.uri();
+    let tmp_path = tmp.path().to_string_lossy().into_owned();
+
+    let (tx, rx) = mpsc::channel::<bool>();
+    let release_watch = release.clone();
+
+    tokio::task::spawn_blocking(move || {
+        install_mock_keyring();
+        let core = std::sync::Arc::new(
+            LyrebirdCore::new(CoreConfig {
+                data_dir: tmp_path,
+                device_name: "Test".into(),
+            })
+            .expect("core init"),
+        );
+        core.login(server_url, "u-861".into(), "pw".into())
+            .expect("login");
+
+        // Watcher: while logout's HTTP is in flight, prove `inner` is
+        // acquirable. It only becomes acquirable if logout dropped the guard
+        // before `block_on`.
+        let core_watch = core.clone();
+        let watcher = std::thread::spawn(move || {
+            let deadline = Instant::now() + Duration::from_secs(5);
+            while Instant::now() < deadline {
+                if core_watch.inner.try_lock().is_some() {
+                    release_watch.store(true, Ordering::SeqCst);
+                    return true;
+                }
+                std::thread::sleep(Duration::from_millis(2));
+            }
+            // Watchdog: never acquired the lock — unblock the server so the
+            // logout thread can return and the test reports the real failure.
+            release_watch.store(true, Ordering::SeqCst);
+            false
+        });
+
+        core.logout().expect("logout");
+        let acquired = watcher.join().unwrap_or(false);
+        tx.send(acquired).ok();
+    });
+
+    let acquired = rx
+        .recv_timeout(Duration::from_secs(10))
+        .expect("logout deadlocked holding Inner across the HTTP round-trip (#861)");
+    assert!(
+        acquired,
+        "watcher could not acquire Inner while logout's HTTP was in flight \
+         — logout held the mutex across block_on (#861)"
+    );
+}
+
 // ============================================================================
 // Playlist CRUD — rename / delete / reorder (#564)
 // ============================================================================
--- a/core/src/tests.rs
+++ b/core/src/tests.rs
@@ -5879,16 +5879,6 @@ async fn logout_tolerates_server_post_failure() {
     .expect("join task");
 }
 
-/// Regression for #861: `logout()` must release the `Inner` mutex *before*
-/// `block_on(post_logout_session())`, matching the `with_client` invariant.
-///
-/// The `/Sessions/Logout` handler parks until a watcher thread can acquire
-/// `core.inner` via `try_lock`. With the bug (guard held across the HTTP
-/// round-trip) the watcher can never take the lock, so the handler is never
-/// released and `logout()` deadlocks; the 10s `recv_timeout` watchdog then
-/// trips and the test fails. With the fix, the guard is dropped before
-/// `block_on`, the watcher takes the lock, signals the handler, and
-/// `logout()` returns promptly.
 #[tokio::test]
 async fn logout_releases_inner_before_http_round_trip() {
     use std::sync::atomic::{AtomicBool, Ordering};
@@ -5909,8 +5899,6 @@ async fn logout_releases_inner_before_http_round_trip() {
         .mount(&server)
         .await;
 
-    // The logout POST parks until it observes `release`, which the watcher
-    // thread only sets once it has proven `core.inner` is acquirable.
     let release = std::sync::Arc::new(AtomicBool::new(false));
     let release_mock = release.clone();
     Mock::given(method("POST"))
@@ -5942,9 +5930,6 @@ async fn logout_releases_inner_before_http_round_trip() {
         core.login(server_url, "u-861".into(), "pw".into())
             .expect("login");
 
-        // Watcher: while logout's HTTP is in flight, prove `inner` is
-        // acquirable. It only becomes acquirable if logout dropped the guard
-        // before `block_on`.
         let core_watch = core.clone();
         let watcher = std::thread::spawn(move || {
             let deadline = Instant::now() + Duration::from_secs(5);
@@ -5955,8 +5940,6 @@ async fn logout_releases_inner_before_http_round_trip() {
                 }
                 std::thread::sleep(Duration::from_millis(2));
             }
-            // Watchdog: never acquired the lock — unblock the server so the
-            // logout thread can return and the test reports the real failure.
             release_watch.store(true, Ordering::SeqCst);
             false
         });


skalthoff:

reviewer-session: wf_cd55d074-0e0-27
diff-stat: 1 file changed, 4 insertions(+), 6 deletions(-)

Checklist:
- [✓ N/A] error-swallowing — diff adds no try?, .ok(), empty-catch, or _ = try patterns. The pre-existing try? in AppModel.logout()'s Task.detached is not introduced by this PR.
- [✓ N/A] MainActor-blocking FFI — diff is pure Rust; no Swift changes at all.
- [✓ N/A] paged-cache-only resolution — no resolution logic touched.
- [✓ N/A] optimistic-UI-without-echo — no UI mutation touched.
- [✓ N/A] hotspot-file growth — only core/src/lib.rs is touched (not a hotspot); diff is net -2 lines.
- [✗ finding] new-branch test coverage — The fix corrects a concurrency hazard (Inner mutex held across block_on). No test was added that fails on origin/main and passes on this branch. The three existing logout tests (logout_calls_sessions_logout_endpoint, logout_tolerates_server_post_failure, logout_clears_user_data) exercise the corrected code path but are single-threaded and cannot distinguish a lock-held-during-HTTP from a lock-released-before-HTTP. Issue #861 itself supplied a failing-test sketch. The PR body makes no claim that reproduction is "exclusively manual." Per the checklist rules, this is a required finding.
- [✓ N/A] speculative scope — every changed line traces directly to the logout() lock-release invariant stated in issue #861. No scope creep.
- [✓ N/A] banned-comments — diff adds zero new comments.

Falsification: The following test would fail on origin/main and pass on this branch:

```rust
#[tokio::test]
async fn logout_releases_inner_before_http() {
    let tmp = tempfile::tempdir().unwrap();
    let server = MockServer::start().await;
    // ... login mock ...
    Mock::given(method("POST"))
        .and(path("/Sessions/Logout"))
        .respond_with(ResponseTemplate::new(204).set_delay(Duration::from_millis(400)))
        .mount(&server)
        .await;

    tokio::task::spawn_blocking(move || {
        let core = Arc::new(JellifyCore::new(CoreConfig {
            data_dir: tmp.path().to_string_lossy().into_owned(),
            device_name: "T".into(),
        }).unwrap());
        core.login(server.uri(), "u".into(), "p".into()).unwrap();

        let core2 = Arc::clone(&core);
        let reader = std::thread::spawn(move || {
            std::thread::sleep(Duration::from_millis(50)); // let logout acquire lock
            let t = std::time::Instant::now();
            let _ = core2.inner.lock().device_id.clone(); // would stall on main
            t.elapsed()
        });
        core.logout().unwrap();
        let elapsed = reader.join().unwrap();
        assert!(
            elapsed < Duration::from_millis(100),
            "Inner lock held during HTTP: device_id() stalled {elapsed:?}"
        );
    }).await.unwrap();
}
```

On origin/main the 400 ms HTTP delay leaks into the reader thread's elapsed time. On this branch, the Arc is cloned and the lock is released before block_on, so elapsed is near-zero.

Outcome: request-changes

Required action: Add a test in core/src/tests.rs that (a) installs a slow mock for POST /Sessions/Logout, (b) concurrently acquires the inner lock from a second thread while logout is in-flight, and (c) asserts the concurrent acquisition completes in well under the server-response latency.

skalthoff:
Addressed the Category 6 (test coverage) finding in 7b133f6.

Added `logout_releases_inner_before_http_round_trip` to `core/src/tests.rs` — a deterministic version of the failing-test sketch from #861. A wiremock `/Sessions/Logout` handler parks on an `AtomicBool` until a watcher thread can acquire `core.inner` via `try_lock` while logout's HTTP round-trip is in flight. Under the pre-fix code the guard is held across `block_on`, so the watcher never takes the lock and `logout()` deadlocks; a 5s watcher deadline plus a 10s `recv_timeout` convert the hang into a clean failure. After the fix the guard is dropped first, the watcher takes the lock, signals the handler, and `logout()` returns.

Verification (both run locally):
- Reverted `logout` to the buggy lock-held-across-`block_on` form -> test FAILED (`test result: FAILED. 0 passed; 1 failed`, tripped the 10s watchdog).
- With the fix on this branch -> PASSED in 0.13s.

Gates green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings` (0 warnings), `cargo test --workspace --exclude lyrebird-desktop --all-features --no-fail-fast` (all suites ok, 0 failed). Test-only change, no FFI surface touched, so no xcframework/bindings regeneration.

skalthoff:

reviewer-session: wf_cd55d074-0e0-31
diff-stat: 3 files changed, +102/-7 (pipeline block claims 1 file, +4/-6 — stale at branch creation)

Checklist:
- [✓ N/A] error-swallowing — `tx.send(acquired).ok()` is test infrastructure; the receiver dropping means the recv_timeout already fired. No production error swallowed.
- [✓ N/A] MainActor-blocking FFI — no Swift changes to AppModel or call-site patterns.
- [✓ N/A] paged-cache-only resolution — no cache-resolution code changed.
- [✓ N/A] optimistic-UI-without-echo — no UI mutation code changed.
- [✗ FINDING] hotspot-file growth — `core/src/tests.rs` is a documented hotspot (+97 lines). Growth is justified by the regression test. However, `macos/Sources/Lyrebird/Screens/AlbumDetailView.swift` is a stale rebase artifact: the branch forked before commit `d573efe` (fix(macos): AlbumDetailView 'More by Artist' uses eager HStack #866) landed on main. That commit changed `LazyHStack → HStack` to fix the macOS 26.4 Apple Silicon use-after-free in `_ArrayBuffer._consumeAndCreateNew`. This PR's diff shows `HStack → LazyHStack` — merging as-is will silently reintroduce a UAF crash that #866 fixed. The branch must be rebased onto current main before this can merge.
- [✓ addressed] new-branch test coverage — `logout_releases_inner_before_http_round_trip` is present. The test correctly uses a watcher thread with `try_lock` to prove the lock is released before `block_on`, and a parking mock handler that won't return until the watcher succeeds.
- [✗ FINDING] speculative scope — the `AlbumDetailView.swift` line change is not in scope for #861. This is a stale-branch artifact (not intentional scope creep), but the effect is the same: merging reverts the crash fix from #866.
- [✓ N/A] banned-comments — `/// Regression for #861:` follows the established project convention (50+ identical patterns in tests.rs).

Falsification: A concurrent-latency test this PR does not cover — prove that a blocked concurrent FFI call (`.lock()` not `.try_lock()`) returns promptly during the logout HTTP window. The existing test proves lock release (binary yes/no) but not that concurrent callers are unblocked in bounded time. In practice `parking_lot::Mutex` FIFO fairness makes this hold by construction, so no real gap exists.

**The core fix is correct. Blocking issue: the stale `AlbumDetailView.swift` rebase artifact will reintroduce the macOS 26.4 UAF crash from #866 if merged without rebasing onto current main.**

Outcome: request-changes

skalthoff:

reviewer-session: wf_0fd057bd-337-8
diff-stat: 3 files changed, 85 insertions(+), 7 deletions(-) — NOTE: the pipeline block's diff-stat says \"1 file changed, +4/-6\", which is wrong. There are 3 files changed.

---

Checklist:

- [✓ N/A] error-swallowing — The Rust change properly logs the error with `tracing::warn!` instead of swallowing it. The Swift `try? core.logout()` at AppModel.swift:810 is pre-existing (not introduced by this PR) and is intentional (fire-and-forget from the MainActor). N/A for this PR.

- [✓ N/A] MainActor-blocking FFI — The Swift side already wraps `core.logout()` in `Task.detached`. The Rust-side fix releases the mutex before `block_on`, which is the invariant being restored. N/A.

- [✓ N/A] paged-cache-only resolution — No new cache lookups introduced. N/A.

- [✓ N/A] optimistic-UI-without-echo — `logout()` is a teardown operation, not a mutation of persisted state. No optimistic UI pattern applies. N/A.

- [✗ FINDING] hotspot-file growth — `core/src/tests.rs` grew by 80 lines (new test `logout_releases_inner_before_http_round_trip`). Growth is justified: the test directly verifies the invariant being restored. HOWEVER — `AlbumDetailView.swift` is not a hotspot file and its change is not mentioned in the PR at all (see speculative-scope finding below). The pipeline block reports a diff-stat that doesn't match the actual diff, which obscures the presence of this third file.

- [✓ addressed] new-branch test coverage — The new `logout_releases_inner_before_http_round_trip` test directly exercises the mutex-release invariant: it installs a mock `/Sessions/Logout` handler that blocks until a concurrent watcher signals it was able to acquire `inner.try_lock()`. On `origin/main` (pre-fix), the watcher would never succeed while the HTTP is in flight; on this branch it will. Coverage is adequate for the Rust fix.

- [✗ FINDING] speculative scope — `macos/Sources/Lyrebird/Screens/AlbumDetailView.swift` is changed: inside `moreByArtistSection`, `HStack` is replaced with `LazyHStack`. This change:
  1. Is not mentioned in the PR title, description, or linked issue #861.
  2. Is entirely unrelated to the mutex lock-release fix in `core/src/lib.rs`.
  3. Reintroduces the **exact pattern that CLAUDE.md rc9 documents as a resolved UAF crash** on macOS 26.4 + Apple Silicon. `ArtistDetailView.swift:493-502` documents the UAF in detail; `GenreDetailView.swift:8-9` explicitly says \"eagerly (no `LazyHStack`) per CLAUDE.md's rc9 note about the macOS 26.4 UAF in `LazyHStack` inside a horizontal ScrollView.\" The `AlbumDetailView` `moreByArtistSection` is the same shape: a `LazyHStack` inside a `ScrollView(.horizontal)`, on the same OS version, with the same `HomeAlbumTile` children. This is a crash regression dressed as an unrelated change. REJECT.

- [✓ N/A] banned-comments — The newly added Rust code (lib.rs change and test) contains no `//` comments explaining what the code does, no task references in `//` comments, and no TODO/FIXME. The assert failure message strings contain `(#861)` but those are runtime strings, not code comments; this is borderline but does not constitute a banned comment in the sense CLAUDE.md defines.

---

Falsification: A failing-test sketch this PR does NOT cover:

```
// AlbumDetailView — moreByArtistSection crash regression
// On macOS 26.4 + Apple Silicon, navigate to an AlbumDetailView for an artist
// with >= 2 albums, then navigate away and back again. The LazyHStack inside
// ScrollView(.horizontal) recycles its child views across navigation. SwiftUI's
// _ArrayBuffer._consumeAndCreateNew triggers a use-after-free in
// HVStack.updateCache on the second render pass.
//
// Expected: no crash.
// With this PR: SIGSEGV inside HVStack.updateCache on the second navigation.
// Without this PR (HStack): no crash.
//
// This is not an edge case; it affects every artist with >= 2 albums.
// The ArtistDetailView fix in rc9 was motivated by exactly this crash shape.
```

No automated test can cover this (SwiftUI framework UAF that only manifests on device), but the code change directly reintroduces a pattern that CLAUDE.md rc9 documents as having caused real crashes.

---

Outcome: request-changes

Required actions before re-review:
1. Revert the `LazyHStack` change in `macos/Sources/Lyrebird/Screens/AlbumDetailView.swift` (restore `HStack`). This change is out of scope for #861 and reintroduces a documented crash regression.
2. Fix the `diff-stat` in the pipeline block to reflect the actual 3-file diff.

skalthoff:
## Fixer response (review-fix pass)

Re-checked both required actions against the **authoritative PR diff** (`gh pr diff 865`, `GET /pulls/865/files`, and `git diff origin/main...origin/fix/861-logout-lock-release`). All three agree the PR touches exactly **2 files**:

- `core/src/lib.rs` (+4/-6)
- `core/src/tests.rs` (+80/-0)

**1. AlbumDetailView LazyHStack revert — no action needed (not in the diff).**
The `HStack -> LazyHStack` change the review flagged only appears when diffing against a *stale* local `main` (`33263ff`) captured before `d573efe` (the #866 eager-HStack fix) landed. Against current `origin/main` (`d573efe`) the file is identical on both sides, so it is not part of this PR and there is nothing to revert. Editing `AlbumDetailView.swift` here would itself be out-of-scope for #861. (An earlier reviewer note already identified this as a "stale rebase artifact" rather than an intentional change — confirmed it fully resolves against current main with no rebase even required, since the merge-base diff is clean.)

**2. Pipeline diff-stat corrected** to `2 files changed, +84/-6`.

**Gates re-run on PR head `5bc5229` (detached worktree):**
- `cargo fmt --all -- --check` — pass
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — pass (0 errors; the sole `unused import` warning is pre-existing in the untouched `linux/` crate, not in PR-changed files)
- `cargo test --workspace --exclude lyrebird-desktop --all-features --no-fail-fast` — 694 passed / 0 failed, incl. `logout_releases_inner_before_http_round_trip ... ok`

No FFI `Record`/`Enum` surface touched, so no binding/xcframework regeneration. No Swift in the diff, so `swift build` is N/A.

skalthoff:

reviewer-session: wf_0fd057bd-337-12
diff-stat: 2 files changed, +84/-6 (core/src/lib.rs, core/src/tests.rs only)

Checklist:
- [N/A] error-swallowing: Preserved if-let-Err + tracing::warn. No swallowing.
- [N/A] MainActor-blocking FFI: No Swift changes introduced by this PR.
- [N/A] paged-cache-only resolution: No list/resolution code touched.
- [N/A] optimistic-UI-without-echo: No UI mutation code.
- [N/A] hotspot-file growth: tests.rs grew 80 lines (justified regression test); lib.rs shrank 6 lines net. AppModel.swift and client.rs untouched.
- [FINDING] new-branch test coverage: The test has a timing race that can produce a false pass on the buggy code. The watcher thread is spawned before core.logout() is called, with no synchronization barrier. On the pre-fix code, if the watcher executes its first try_lock() before logout() acquires inner.lock(), it acquires the free mutex, sets release=true (unblocking the mock HTTP), returns true, and the test passes spuriously. Thread creation latency makes this improbable in practice, but it is not deterministic. Fix: wait for wiremock to have received the POST /Sessions/Logout request before the watcher begins spinning, ensuring logout() is definitely in flight at the block_on boundary when try_lock() is first attempted.
- [N/A] speculative scope: Every changed line traces to issue #861. No unrelated modifications.
- [N/A] banned-comments: No code comments added. #861 references appear only in expect()/assert!() string literals.

Falsification (uncovered case):
Concurrent double-logout (user double-clicks sign-out while the first logout HTTP is in flight). Both calls clone the Arc successfully, both fire POST /Sessions/Logout, the second gets 401 (session already gone), the tracing::warn path absorbs it. Behavior is correct but untested. Not a blocker by itself.

The primary finding is the test timing race described under new-branch test coverage. The watcher has no guarantee it observes logout() holding the mutex at the time of its first try_lock() attempt.

Outcome: request-changes

skalthoff:

reviewer-session: wf_cd0923b5-888-3
diff-stat: 2 files changed, +84/-6 (lib.rs net -2, tests.rs +80)

Checklist:
- [✓ N/A] error-swallowing — `tx.send(acquired).ok()` is test-only; channel send fails only when receiver is already dropped (timeout path). Production error path unchanged: warn-and-continue for failed POST /Sessions/Logout is pre-existing and appropriate per #592 rationale.
- [✓ N/A] MainActor-blocking FFI — pure Rust change; no Swift files modified.
- [✓ N/A] paged-cache-only resolution — no data resolution code touched.
- [✓ N/A] optimistic-UI-without-echo — no UI mutations.
- [✓ N/A] hotspot-file growth — tests.rs grew +80 lines, justified as regression test for this exact fix. lib.rs is not a hotspot and shrank net 2 lines.
- [✓ addressed] new-branch test coverage — `logout_releases_inner_before_http_round_trip` correctly blocks the mock HTTP handler on an AtomicBool until the watcher proves `inner.try_lock()` succeeds during the in-flight request, then unblocks logout. Fails on main (bug), passes after fix.
- [✓ N/A] speculative scope — only `logout()` in lib.rs modified; change is a direct surgical application of the `with_client` invariant. No other functions touched.
- [✓ N/A] banned-comments — no `//` comments added. `#861` references appear in `assert!` / `.expect()` panic strings (runtime diagnostic output, not source comments). No TODO/FIXME added.

Falsification: A concurrent double-logout race (two threads both call `logout()` simultaneously after the fix) could issue POST /Sessions/Logout twice. Neither the test nor the production code guards against this. However, no such invariant is documented, and double-logout is outside the scope of #861 (which only addresses the lock-held-across-block_on violation). The fix correctly implements the documented `with_client` invariant: clone Arc under brief lock, drop guard, then run HTTP. No falsifying case exists for the stated invariant.

Outcome: approve

PR #896, opened 2026-06-01
macos: add output device picker in Preferences with AudioEngine routing

### LINKED ISSUE #262: Playback — output device picker
- Lists Core Audio devices via `AVAudioSession`. Persists selection + falls back gracefully.
- "Exclusive mode" toggle for lossless output.

<sub>Source: `06-screen-specs.md`</sub>

skalthoff:
Adds a Core Audio output-device picker (+ exclusive/hog mode toggle) to the Audio preferences pane so playback can be pinned to a specific device instead of always following the system default.

Closes #262

### What changed
- `LyrebirdAudio/AudioOutputDevices.swift` — HAL enumeration of output-capable devices (stable `kAudioDevicePropertyDeviceUID`), UID→device resolution, and hog-mode acquire/release.
- `AudioEngine` — new `outputDeviceUID` property pins `AVPlayer.audioOutputDeviceUniqueID` on the live player and every player it builds (play / gapless preload reuse / stall recovery). Empty/unknown UID = system default (the macOS-correct knob; `AVAudioSession` is iOS-only).
- `AppModel` — seeds the engine from persisted storage at startup; `setOutputDevice`/`setExclusiveMode` route changes off the `MainActor`, surface `errorMessage`, and roll back the toggle on failure (matching the optimistic-UI-without-echo pattern). Switching to **System Default** clears the `exclusiveMode` flag and releases the old hog claim, so a stale flag can't silently hog-claim the next concrete device.
- `PreferencesAudio` — Output Device section with a "System Default" option, a stale-device fallback hint, and an Exclusive Mode switch (disabled until a concrete device is chosen).

### Notes
- Device enumeration runs in a detached task (HAL reads can block the main actor).
- The pane now reads `Environment(AppModel.self)`; its bare `#Preview` was removed for the same reason `PreferencesServer`/`PreferencesAccount` omit theirs.

`pipeline:`
- fixer-session: feat/262-output-device-picker
- slice: audio
- hotspots-claimed: AppModel.swift (+85; output-device routing methods)
- build-gate: pass (clean `swift build`; `AudioOutputDevicesTests` 5/5 green; no Rust source changed, so no FFI/binding regeneration needed)
- diff-stat: 5 files, +450/-7 (2 new files)

### DIFF
--- a/macos/Sources/Lyrebird/AppModel.swift
+++ b/macos/Sources/Lyrebird/AppModel.swift
@@ -841,6 +841,11 @@ final class AppModel {
         self.mediaSession.attach(delegate: self)
         self.audio.mediaSession = self.mediaSession
         self.audio.delegate = self
+        // Seed the engine with the persisted output-device selection (#262)
+        // so the first track honours it without waiting for the Preferences
+        // pane to mount. An empty/absent value means "follow system default".
+        let savedDeviceUID = UserDefaults.standard.string(forKey: AudioOutputDevices.preferenceKey) ?? ""
+        self.audio.outputDeviceUID = savedDeviceUID.isEmpty ? nil : savedDeviceUID
     }
 
     // MARK: - Network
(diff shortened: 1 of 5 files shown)


skalthoff:
Addressed all four findings on the same branch (`feat/262-output-device-picker`):

1. **Optimistic-UI bug (System Default + exclusive)** — `setOutputDevice` now special-cases an empty UID: it clears `exclusiveModePreferenceKey` and releases the old device's hog claim, instead of leaving the flag `true` (which previously caused the next concrete device pick to silently hog-claim). The `setExclusiveMode(_:forUID:"")` no-op contract is unchanged; the fix is at the caller so the persisted flag matches reality.

2. **Diff-stat / hotspots** — the AppDelegate/MediaSession changes are gone (rebased onto current main, which already carries #893/#894), so the diff is now a true 5 files / +450 / -7 (2 new files). PR body corrected, and it now declares the `AppModel.swift` hotspot touch explicitly (output-device routing methods) rather than "none".

3. **Speculative scope** — no longer present. `AppDelegate.swift` (typed-enum vs string match) and `MediaSession.swift` (NSCache cap) are not in this diff; both were unrelated and landed via their own merged PRs.

4. **Banned what-does comments** — trimmed the narration comments in `AudioOutputDevices.swift` (`device(forUID:)` doc), `PreferencesAudio.swift` (both binding docs), and the `AudioEngine` play-site inline comment to why-only.

Build gate: clean `swift build` (FFI-adjacent clean rebuild), `AudioOutputDevicesTests` 5/5 green. No Rust source changed, so no binding/xcframework regeneration was required.

PR #898, opened 2026-06-01
fix(core): widen bitrate to i64 so PlaybackInfo survives the -1000 sentinel

### LINKED ISSUE #890: core: MediaSourceInfo.bitrate typed as u32 silently breaks playback session tracking for FLAC tracks with Bitrate: -1000
## Summary

`MediaSourceInfo.bitrate` is typed `Option<u32>` in `core/src/models.rs:544`. The Jellyfin server returns `\"Bitrate\": -1000` for some FLAC tracks in the `POST /Items/{id}/PlaybackInfo` response. When serde encounters `-1000` and the target type is `u32`, deserialization of the entire `PlaybackInfoResponse` fails — `#[serde(default)]` only substitutes when the field is *absent*, not when the field is present with a wrong type.

## File + Line Citations

**`core/src/models.rs:537–561`** — `MediaSourceInfo.bitrate` declared as `Option<u32>`:
```rust
#[derive(Clone, Debug, Default, Serialize, Deserialize, uniffi::Record)]
#[serde(rename_all = "PascalCase")]
pub struct MediaSourceInfo {
    #[serde(default)]
    pub id: String,
    #[serde(default)]
    pub path: Option<String>,
    #[serde(default)]
    pub container: Option<String>,
    #[serde(default)]
    pub bitrate: Option<u32>,   // ← fails on -1000 from server
    ...
}
```

**`macos/Sources/LyrebirdAudio/AudioEngine.swift:159`** — error swallowed with `try?`:
```swift
guard let info = try? core.playbackInfo(itemId: trackId, opts: opts) else {
    return (nil, nil)  // ← returns when -1000 causes deserialization error
}
```

The comment at `AudioEngine.swift:143–146` says nil is "harmless … URL is still playable." That is true for audio delivery, but nil `playSessionId` means:
- "Now Playing on macOS" does not appear on other Jellyfin clients for affected tracks
- `PlayCount` and resume-position ticks are not updated server-side

## Reproduction

```bash
TOKEN=$(curl -sS -X POST "https://music.skalthoff.com/Users/AuthenticateByName" \
  -H "Content-Type: application/json" \
  -H 'Authorization: MediaBrowser Client="probe", Device="probe", DeviceId="probe-001", Version="0.0.0"' \
  -d '{"Username":"test","Pw":"test"}' | python3 -c 'import json,sys; d=json.load(sys.stdin); print(d["AccessToken"] + "|" + d["User"]["Id"])')
ACCESS_TOKEN=$(echo $TOKEN | cut -d'|' -f1)
USER_ID=$(echo $TOKEN | cut -d'|' -f2)

# Known problematic track id
curl -sS -X POST "https://music.skalthoff.com/Items/b6a8a7dc72d24083535a6fe7add04527/PlaybackInfo?userId=$USER_ID" \
  -H "Content-Type: application/json" \
  -H "Authorization: MediaBrowser Token=\"$ACCESS_TOKEN\"" \
  -d '{}' | python3 -c 'import json,sys; d=json.load(sys.stdin); print([s.get("Bitrate") for s in d.get("MediaSources",[])])'
# Expected output: [-1000]  (a negative integer — fails serde as u32)
```

Approximately 1 in 10 sampled FLAC tracks on the test server returns `Bitrate: -1000`. Jellyfin uses `-1000` as a sentinel meaning "bitrate unknown/not measurable" (seen with losslessly-encoded tracks where file analysis failed). The correct fix is to change `bitrate: Option<u32>` → `bitrate: Option<i64>` in `MediaSourceInfo` (and the same field in `RawItem` at `core/src/client.rs:2605`), or use `#[serde(deserialize_with = ...)]` to treat negative values as `None`.

## Falsifiability

This is wrong because serde's `#[serde(default)]` only activates on a missing JSON field, not on a present-but-type-incompatible value; so parsing `{\"Bitrate\": -1000}` into `Option<u32>` returns an error. If this claim is wrong, this issue is invalid.

## De-dup

De-dup checked against open issues (search queries: `MediaSourceInfo bitrate`, `FLAC bitrate playback`, `PlaybackInfo deserialize`, `bitrate u32`) and CLAUDE.md Deferred list. No match.

skalthoff:
Jellyfin emits `Bitrate: -1000` as an "unmeasurable" sentinel for some FLAC tracks, and `Option<u32>` rejected it during deserialization — sinking the whole `PlaybackInfo` response (and any item listing carrying the sentinel), which silently disabled playback-session tracking, `PlayCount`, and Now-Playing-on-other-clients for those tracks.

This widens the `bitrate` field to `Option<i64>` on `MediaSourceInfo` and `ItemRef` (and the internal `RawItem`) so the negative sentinel deserializes cleanly; downstream display code already guards `bitrate > 0`, so sentinels are filtered at render time exactly as before.

Build-gate notes:
- Rust: `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -D warnings`, `cargo test --workspace --exclude lyrebird-desktop --all-features` all pass (incl. new regression test `playback_info_parses_negative_bitrate_sentinel`).
- Swift: clean `rm -rf macos/.build && swift build --package-path macos` after regenerating the xcframework + bindings (these are gitignored / CI-regenerated in this repo, so not committed). The `Int64?` migration required updating `formatBitrate(_:)` in `NowPlayingView` and the DTO parser's `bitrate` local in `AppModel` to match the new binding type.

Hotspot lock claimed: #892 (`clientrs`) — release on merge.

Closes #890

pipeline:
  fixer-session: wf_367922fa-3ba-34
  slice: slice:models
  hotspots-claimed: [clientrs]
  build-gate: pass
  diff-stat: 5 files changed, +47/-6

### DIFF
--- a/core/src/client.rs
+++ b/core/src/client.rs
@@ -2632,7 +2632,7 @@ pub struct RawItem {
     #[serde(rename = "Container")]
     pub container: Option<String>,
     #[serde(rename = "Bitrate")]
-    pub bitrate: Option<u32>,
+    pub bitrate: Option<i64>,
     #[serde(rename = "Path")]
     pub path: Option<String>,
     #[serde(rename = "PlaylistItemId")]
--- a/core/src/models.rs
+++ b/core/src/models.rs
@@ -76,7 +76,7 @@ pub struct Track {
     /// See [`Self::is_favorite`].
     pub play_count: u32,
     pub container: Option<String>,
-    pub bitrate: Option<u32>,
+    pub bitrate: Option<i64>,
     pub image_tag: Option<String>,
     /// Jellyfin `PlaylistItemId` — only populated when the track was fetched
     /// as part of a playlist (via `playlist_tracks`). Used by
@@ -545,7 +545,7 @@ pub struct MediaSourceInfo {
     #[serde(default)]
     pub container: Option<String>,
     #[serde(default)]
-    pub bitrate: Option<u32>,
+    pub bitrate: Option<i64>,
     #[serde(default)]
     pub size: Option<i64>,
     #[serde(default)]
--- a/core/src/tests.rs
+++ b/core/src/tests.rs
@@ -7778,3 +7778,36 @@ async fn playback_info_parses_negative_bitrate_sentinel() {
     assert_eq!(resp.media_sources.len(), 1);
     assert_eq!(resp.media_sources[0].bitrate, Some(-1000));
 }
+
+#[tokio::test]
+async fn album_tracks_parses_negative_bitrate_sentinel() {
+    let server = MockServer::start().await;
+    Mock::given(method("POST"))
+        .and(path("/Users/AuthenticateByName"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "AccessToken": "t", "ServerId": "s", "ServerName": "S",
+            "User": { "Id": "u1", "Name": "user", "ServerId": "s", "PrimaryImageTag": null }
+        })))
+        .mount(&server)
+        .await;
+    Mock::given(method("GET"))
+        .and(path("/Users/u1/Items"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "Items": [
+                {
+                    "Id": "t1", "Name": "Yona", "Type": "Audio",
+                    "AlbumId": "a1", "Container": "flac", "Bitrate": -1000,
+                    "RunTimeTicks": 2220000000u64
+                }
+            ],
+            "TotalRecordCount": 1
+        })))
+        .mount(&server)
+        .await;
+
+    let mut client = mock_client(&server.uri());
+    client.authenticate_by_name("user", "pw").await.unwrap();
+    let tracks = client.album_tracks("a1").await.unwrap();
+    assert_eq!(tracks.len(), 1);
+    assert_eq!(tracks[0].bitrate, Some(-1000));
+}
--- a/core/src/tests.rs
+++ b/core/src/tests.rs
@@ -7737,3 +7737,44 @@ async fn suggestions_rejects_non_audio_items() {
     assert_eq!(tracks[0].id, "t1");
     assert_eq!(tracks[0].name, "Real Track");
 }
+
+#[tokio::test]
+async fn playback_info_parses_negative_bitrate_sentinel() {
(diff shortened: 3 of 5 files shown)


skalthoff:

reviewer-session: wf_367922fa-3ba-59
diff-stat: 5 files changed, +80/-6 (core/src/client.rs +1/-1, core/src/models.rs +2/-2, core/src/tests.rs +74, macos/.../AppModel.swift +2/-2, macos/.../NowPlayingView.swift +1/-1)

Checklist:
- [✓ N/A] error-swallowing — no try?, .ok(), _ = try, or empty catch added
- [✓ N/A] MainActor-blocking FFI — no try core.X(...) / try await core.X(...) added; pure type annotation change
- [✓ N/A] paged-cache-only resolution — no .first { $0.id == X } pattern added
- [✓ N/A] optimistic-UI-without-echo — no local mutation added; read-side type change only
- [✓ N/A] hotspot-file growth — client.rs +0 net (1 line changed in place); tests.rs +74 lines but all test coverage; AppModel.swift +0 net; NowPlayingView.swift +0 net
- [✓ addressed] new-branch test coverage — two Rust tests added (playback_info_parses_negative_bitrate_sentinel, album_tracks_parses_negative_bitrate_sentinel) covering both affected deserialization paths; both fail on main (Option<u32> rejects -1000), pass on this branch
- [✓ N/A] speculative scope — every changed line traces directly to #890 (negative bitrate sentinel sinking PlaybackInfo/album-tracks deserialization); no unrelated lines
- [✓ N/A] banned-comments — no comments added in the diff at all

Falsification: A `playlist_tracks` path test with Bitrate: -1000 would fail on main and pass on this branch. However, `playlist_tracks` calls `RawItems<RawItem>::json()` → `Track::from(RawItem)` — the exact same `RawItem.bitrate` field widened in this PR. The fix covers it structurally without a dedicated test. No such uncovered case exists because every track-fetching call site in the codebase goes through `RawItem → Track` conversion, and `RawItem.bitrate` is the single field widened.

Swift callers not updated (TrackInfoSheet, FormatBadge, AlbumDetailView) all use type-inferred `let bitrate = track.bitrate` / `$0.bitrate` patterns with `> 0` guards — fully compatible with `Int64?`. The generated Swift bindings are gitignored and CI-regenerated from source.

Outcome: approve

PR #942, opened 2026-06-02
fix(macos): CommandPalette shows raw id for Go to Favorites action

### LINKED ISSUE #937: fix(macos): CommandPalette shows raw id string for Go to Favorites action
## Summary

The \`nav.favorites\` palette action registered in \`AppModel.paletteActions\` (AppModel.swift:579) is missing from the two static lookup tables in \`PaletteRow\` inside \`CommandPalette.swift\`. As a result, when a user opens the ⌘K command palette, the "Go to Favorites" action renders with the raw id string \`nav.favorites\` as its label and a generic bolt icon instead of its intended title/icon.

## Affected code

**AppModel.swift line 579–583** — action is registered:

\`\`\`swift
actions.append(PaletteAction(
    id: "nav.favorites",
    title: "Go to Favorites",
    symbol: "heart",
    run: { [weak self] in self?.selectTab(.favorites) }
))
\`\`\`

**CommandPalette.swift lines 590–618** — \`actionTitleById\` and \`actionSymbolById\` static maps are missing the key:

\`\`\`swift
// "nav.favorites" is absent here:
static let actionTitleById: [String: String] = [
    "nav.library": "Go to Library",
    "nav.home": "Go to Home",
    "nav.discover": "Go to Discover",
    // nav.favorites: ← missing
    ...
]
// "nav.favorites" is absent here too:
static let actionSymbolById: [String: String] = [
    "nav.home": "house",
    "nav.discover": "sparkles",
    // nav.favorites: ← missing
    ...
]
\`\`\`

**CommandPalette.swift lines 551–555** — fallback path renders raw id:

\`\`\`swift
private var primaryText: String {
    switch row {
    case .action(let id): return PaletteRow.actionTitleById[id] ?? id  // ← returns "nav.favorites"
    ...
}
\`\`\`

**CommandPalette.swift lines 543–547** — fallback path renders wrong icon:

\`\`\`swift
private var iconName: String {
    switch row {
    case .action(let id):
        return PaletteRow.actionSymbolById[id] ?? "bolt"  // ← returns "bolt"
    ...
}
\`\`\`

## Reproduction

1. Launch the app against any Jellyfin server with at least one favorited track.
2. Press ⌘K to open the command palette.
3. Scroll down to or type "nav" in the search field.
4. Observe: a row with the label \`nav.favorites\` and a bolt icon appears instead of "Go to Favorites" with a heart icon.

**Swift test sketch:**

\`\`\`swift
// Falsifiability: every id emitted by AppModel.paletteActions must appear in PaletteRow.actionTitleById.
// This test would catch the regression:
let allIds = Set(appModel.paletteActions.map(\.id))
let mappedIds = Set(PaletteRow.actionTitleById.keys)
XCTAssert(allIds.isSubset(of: mappedIds), "Missing: \(allIds.subtracting(mappedIds))")
\`\`\`

## Falsifiability statement

This is wrong because \`PaletteRow.actionTitleById\` / \`actionSymbolById\` do not contain the key \`"nav.favorites"\` while \`AppModel.paletteActions\` emits a \`PaletteAction\` with that id — the fallback \`?? id\` on line 555 and \`?? "bolt"\` on line 546 are therefore reached for that action. If either of those key-miss fallbacks are never reached in practice (i.e., the action is never shown), this issue is invalid.

## Fix

Add the missing entries:

\`\`\`swift
// In PaletteRow.actionTitleById:
"nav.favorites": "Go to Favorites",

// In PaletteRow.actionSymbolById:
"nav.favorites": "heart",
\`\`\`

## De-dup

De-dup checked against open issues (search queries: "nav.favorites", "favorites palette", "command palette", "palette action") and CLAUDE.md Deferred list. No match.

skalthoff:
The Go to Favorites command-palette action registered in `AppModel.paletteActions` was unreachable in the two static lookup tables in `PaletteRow`, so the ⌘K palette rendered the raw id `nav.favorites` with a generic bolt icon instead of its title and heart symbol.

Adds the missing `nav.favorites` key to `actionTitleById` ("Go to Favorites") and `actionSymbolById` ("heart"), matching the registration in `AppModel`. No behavior change to any other action.

Closes #937

pipeline:
    fixer-session: wf_7dda2c5e-608-13
    slice: slice:components
    hotspots-claimed: []
    build-gate: pass
    diff-stat: 1 file changed, +2/-0


skalthoff:

reviewer-session: wf_7dda2c5e-608-33
diff-stat: 2 files changed, 41 insertions(+), 1 deletion(-)

Checklist:
- [N/A] error-swallowing — No try?, empty catch, _ = try, or .ok() in added lines.
- [N/A] MainActor-blocking FFI — No FFI calls introduced in the diff.
- [N/A] paged-cache-only resolution — No .first { $0.id == X } over model collections added.
- [N/A] optimistic-UI-without-echo — No local state mutation. Fix is read-only lookup table entries.
- [N/A] hotspot-file growth — AppModel.swift, LyrebirdApp.swift, client.rs, tests.rs untouched. CommandPalette.swift is not a hotspot.
- [finding] new-branch test coverage — Test is added. Structural gap: AppModel.paletteActions conditionally gates actions (e.g. download.current under supportsDownloads=false). The test constructs AppModel() with defaults so conditionally-gated ids never appear in actionIds. Coverage claim overstates what is checked. Acknowledge or fix.
- [N/A] speculative scope — Two dictionary entries + private->struct visibility change, all traceable to the issue fix.
- [finding] banned-comments — PaletteRowCoverageTests.swift doc comment line 11: "exactly the nav.favorites regression in #937". Per CLAUDE.md, comments referencing the current task by issue number are banned. Remove the #937 reference.

Falsification: a future conditional palette action id missing from the lookup tables (guarded by a flag defaulting to false) would not be caught by this test — it would pass while the raw id is rendered at runtime.

Outcome: request-changes

PR #943, opened 2026-06-02
fix(macos): breadcrumb resolves album/artist names past the paged cache

### LINKED ISSUE #938: MainShell breadcrumb shows ellipsis for albums/artists not in the paged cache
### Summary

`MainShell`'s breadcrumb builder resolves the album and artist name for the current drill destination using `model.albums.first` / `model.artists.first` — a cache-only scan of the locally-loaded first page. On a library with 20k+ albums, only the first ~100 are in `model.albums` at any given time. When the user navigates to an album that was NOT loaded in the first page (e.g. via context menu from the Recent Played list, or via a deep link from `ArtistDetailView.moreByArtistSection`), the breadcrumb falls back to "…" instead of showing the album name.

### Affected code

`macos/Sources/Lyrebird/Screens/MainShell.swift`, lines 313–329:

```swift
case .album(let id)?:
    if model.screen != .library { segments.append("Library") }
    segments.append("Albums")
    if let album = model.albums.first(where: { $0.id == id }) {
        segments.append(album.name)
    } else {
        // Ellipsis is more informative than a section-name-matching
        // literal fallback ("Albums > Album") that reads like a bug.
        segments.append("…")
    }
case .artist(let id)?:
    if model.screen != .library { segments.append("Library") }
    segments.append("Artists")
    if let artist = model.artists.first(where: { $0.id == id }) {
        segments.append(artist.name)
    } else {
        segments.append("…")
    }
```

### Reproduction

1. Connect to a Jellyfin server with >100 albums (e.g. `https://music.skalthoff.com`, 20k+ albums).
2. On the Home screen, navigate from "Recently Played" or "Jump Back In" to an album that is not in the library's alphabetical first page.
3. Observe breadcrumb: reads "Lyrebird > Home > Library > Albums > …" instead of "… > Albums > <Album Name>".

Alternatively:
- Open `AlbumDetailView` for any album. In the "More by Artist" carousel, tap an album. Breadcrumb shows "…" for the second album if it was not in `model.albums`.

### Falsifiability statement

This is wrong because `model.albums.first(where: { $0.id == id })` only scans the in-memory page (typically ~100 entries), so any album navigated to from a source outside `model.albums` — recently played, artist discography, genre detail — produces "…" in the breadcrumb trail. If `model.albums` is always fully populated before navigation happens, this issue is invalid.

### Fix

Look up the name from the `resolveAlbum` / `resolveArtist` cache before falling through to "…". The `AppModel.resolveAlbum(id:)` method already performs a cache-miss FFI lookup and is used by `AlbumDetailView` and `NowPlayingView` for the same purpose. The breadcrumb builder could maintain a lightweight `[String: String]` id→name cache seeded by `resolveAlbum`'s results, or simply use the `fetchedAlbum` record already resolved by the destination view.

A simpler approach: use `model.albumNameCache[id] ?? "…"` where `albumNameCache` is a dictionary that `resolveAlbum` writes into on resolution.

### De-dup checked

De-dup checked against open issues (search query: `breadcrumb`) and CLAUDE.md Deferred list. No match.

skalthoff:
The breadcrumb trail dead-ended at "…" for any album or artist a user drilled into from outside the library's loaded first page (recently played, artist discography, genre detail), which is the common case on a 20k+-album library.

The breadcrumb builder resolved names via `model.albums.first` / `model.artists.first` — a cache-only scan of the in-memory page. This seeds a lightweight id→name cache in `resolveAlbum` / `resolveArtist` (already called on drill-in by `AlbumDetailView` / `ArtistDetailView`) and reads it as a fallback before the ellipsis. The cache is cleared on `logout()` / `forgetToken()`.

Closes #938

pipeline:
  fixer-session: wf_7dda2c5e-608-15
  slice: slice:screens
  hotspots-claimed: []
  build-gate: pass
  diff-stat: 2 files changed, +24/-3

### DIFF
--- a/macos/Sources/Lyrebird/AppModel.swift
+++ b/macos/Sources/Lyrebird/AppModel.swift
@@ -2429,6 +2429,18 @@ final class AppModel {
     /// its name instead of "…". Cleared on `logout()` / `forgetToken()`.
     var resolvedNameCache: [String: String] = [:]
 
+    /// Breadcrumb display name for an album id: the loaded `albums` page first,
+    /// then `resolvedNameCache` (seeded by `resolveAlbum` on drill-in), then nil
+    /// when neither knows the name so the caller can render an ellipsis.
+    func breadcrumbAlbumName(id: String) -> String? {
+        albums.first(where: { $0.id == id })?.name ?? resolvedNameCache[id]
+    }
+
+    /// Breadcrumb display name for an artist id, mirroring `breadcrumbAlbumName`.
+    func breadcrumbArtistName(id: String) -> String? {
+        artists.first(where: { $0.id == id })?.name ?? resolvedNameCache[id]
+    }
+
     /// Resolve an `Album` record by id — cache-first, falling back to
     /// `core.fetchItem` for libraries larger than the loaded `albums`
     /// page. Returns nil on error or missing id.
--- a/macos/Sources/Lyrebird/AppModel.swift
+++ b/macos/Sources/Lyrebird/AppModel.swift
@@ -1029,6 +1029,7 @@ final class AppModel {
         artistPlaylistsCache = [:]
         artistAlbumsCache = [:]
         artistDetailCache = [:]
+        resolvedNameCache = [:]
         recentlyPlayed = []
         forYou = []
         genresToExplore = []
@@ -1096,6 +1097,7 @@ final class AppModel {
         artistPlaylistsCache = [:]
         artistAlbumsCache = [:]
         artistDetailCache = [:]
+        resolvedNameCache = [:]
         recentlyPlayed = []
         forYou = []
         genresToExplore = []
@@ -2377,8 +2379,12 @@ final class AppModel {
     /// "N albums · M songs" strip silently hides the zero-count lines
     /// rather than lying.
     func resolveArtist(id: String) async -> Artist? {
-        if let cached = artists.first(where: { $0.id == id }) { return cached }
+        if let cached = artists.first(where: { $0.id == id }) {
+            resolvedNameCache[id] = cached.name
+            return cached
+        }
         guard let detail = await artistDetail(artistId: id) else { return nil }
+        resolvedNameCache[id] = detail.name
         return Artist(
             id: detail.id,
             name: detail.name,
@@ -2416,6 +2422,13 @@ final class AppModel {
     /// Per-session cache for `artistDetail`. Cleared on `logout()` / `forgetToken()`.
     private var artistDetailCache: [String: ArtistDetail] = [:]
 
+    /// Album/artist id → display name, seeded by `resolveAlbum` / `resolveArtist`
+    /// when they resolve an id past the loaded library page. The breadcrumb
+    /// builder reads this so a drill destination reached from outside
+    /// `albums` / `artists` (recently played, discography, genre detail) shows
+    /// its name instead of "…". Cleared on `logout()` / `forgetToken()`.
+    var resolvedNameCache: [String: String] = [:]
+
     /// Resolve an `Album` record by id — cache-first, falling back to
     /// `core.fetchItem` for libraries larger than the loaded `albums`
     /// page. Returns nil on error or missing id.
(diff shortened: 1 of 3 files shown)


skalthoff:

reviewer-session: wf_7dda2c5e-608-23
diff-stat: 2 files changed, +24/-3 (AppModel.swift +18, MainShell.swift +6)

Checklist:
- [✓ N/A] error-swallowing — No new `try?`, empty `catch`, or `_ = try` in the diff. The existing `catch` blocks in `resolveAlbum`/`resolveArtist` call `handleAuthError(error)` and return nil — pre-existing pattern, not introduced by this PR.
- [✓ N/A] MainActor-blocking FFI — `breadcrumbSegments` is a synchronous computed property reading from an in-memory `[String: String]` dictionary — zero FFI calls. The seeding paths in `resolveArtist`/`resolveAlbum` are `async` and already use `Task.detached` for FFI. Clean.
- [✓ N/A] paged-cache-only resolution — The PR exists to fix this pattern. The breadcrumb builder now has a three-tier lookup (in-page cache → `resolvedNameCache` → "…") and the name cache is seeded by the existing `resolveAlbum`/`resolveArtist` which fall through to FFI on miss.
- [✓ N/A] optimistic-UI-without-echo — No mutations in this PR.
- [✓ N/A] hotspot-file growth — AppModel.swift grows 18 lines (5937 → 5955). Minimum necessary: `resolvedNameCache` declaration + docstring (~7 lines) + two-line seeding additions in the two existing resolvers. The cache must live on the same `@MainActor` object as the resolvers; no alternative home exists.
- [✗ finding] new-branch test coverage — No test was added. The fix is behaviorally testable: inject an `AppModel` with empty `albums` and a pre-seeded `resolvedNameCache` entry, verify `breadcrumbSegments` returns the cached name instead of "…". The project has an active Swift test suite (`macos/Tests/LyrebirdTests/`). This path is missing.
- [✓ N/A] speculative scope — Every changed line traces to issue #938 (breadcrumb shows "…" for out-of-page items). No scope creep.
- [✗ finding] banned-comments — The inline comment added in `MainShell.swift` lines 372–373 ("The album was navigated to from outside the loaded library page; its name was seeded into the cache by `resolveAlbum`.") explains *what* the code does, which the code already makes visible. CLAUDE.md explicitly bans "Comments explaining what code does (the code already shows that)." The `///` docstring for `resolvedNameCache` in AppModel.swift is acceptable (it documents contract/lifecycle, not implementation).

Falsification:

```swift
// Hypothetical BreadcrumbResolvedNameCacheTests.swift (not present in PR)
func testBreadcrumbShowsNameForAlbumNotInLibraryPage() {
    let model = AppModel(/* test init */)
    model.albums = []  // album not in the loaded page
    model.resolvedNameCache["album-xyz"] = "Dark Side of the Moon"
    model.navPath = [.album("album-xyz")]
    let shell = MainShell()
    // Act: read breadcrumbSegments via model environment
    let segments = shell.breadcrumbSegments
    // Fails on origin/main ("…"), passes on this branch ("Dark Side of the Moon")
    XCTAssertEqual(segments.last, "Dark Side of the Moon")
}
```

Required fixes before merge:
1. Remove the inline `//` comment on lines 372–373 in `MainShell.swift` (banned: explains what, not why).
2. Add a test in `macos/Tests/LyrebirdTests/` that exercises the cache-fallback path in `breadcrumbSegments` — fails on main, passes on branch. If `AppModel`'s initializer cannot be constructed in the test target without a live core, state that explicitly; "reproduction is exclusively manual" requires stating it, not just omitting the test.

Outcome: request-changes

skalthoff:
Addressed both review findings:

1. Removed the inline comment restating what the fallback branch does in `MainShell.swift`.
2. Added test coverage. The breadcrumb name resolution (page cache → `resolvedNameCache` → ellipsis) was buried in a private `var` on the `MainShell` view; pulled it into `AppModel.breadcrumbAlbumName(id:)` / `breadcrumbArtistName(id:)` so it is reachable headlessly, and `BreadcrumbNameResolutionTests` exercises all three branches for both album and artist.

Build gate: `swift build` clean; `swift test --filter BreadcrumbNameResolutionTests` → 6 passed.

PR #949, opened 2026-06-02
test(macos): cover capability-gated palette ids in PaletteRow coverage gate

skalthoff:
Follow-up to the palette-favorites fix: the coverage gate only asserted against the default `AppModel` roster, which omits capability-gated actions, so the gated `download.current` row's map entries were never verified; this closes that gap and drops a stale task-number reference from the doc comment.

PR #942 (the original `nav.favorites` fix) squash-merged to main before the reviewer's requested changes could land on its branch, so this carries those two changes forward.

Addresses reviewer findings on #942:
- doc comment no longer references a task number by id
- gate now unions the capability-gated action ids (`download.current`) into the asserted set, so their `actionTitleById`/`actionSymbolById` entries are actually covered regardless of the runtime `supportsDownloads` gate

pipeline:
  fixer-session: wf_7dda2c5e-608-40
  slice: slice:components
  hotspots-claimed: []
  build-gate: pass
  diff-stat: 1 file changed, +12/-3

### DIFF
--- a/macos/Tests/LyrebirdTests/PaletteRowCoverageTests.swift
+++ b/macos/Tests/LyrebirdTests/PaletteRowCoverageTests.swift
@@ -7,11 +7,19 @@ import XCTest
 /// `PaletteRow` labels and icons each action by id through two hand-maintained
 /// static maps (`actionTitleById` / `actionSymbolById`), kept in sync with the
 /// roster emitted by `AppModel.paletteActions`. A drift between the two renders
-/// the raw id string and a generic bolt icon (the key-miss fallbacks) — exactly
-/// the `nav.favorites` regression in #937. This gate fails if any registered
-/// action id is missing from either map.
+/// the raw id string and a generic bolt icon (the key-miss fallbacks) — a
+/// `nav.favorites` row showing its raw id is the regression this guards. This
+/// gate fails if any registered action id is missing from either map.
 @MainActor
 final class PaletteRowCoverageTests: XCTestCase {
+
+	/// Action ids that `paletteActions` only appends behind a capability flag
+	/// (`supportsDownloads`, etc.), so the default `AppModel` roster never
+	/// surfaces them. They still need map entries because the row renders the
+	/// same way once the flag flips, so cover them explicitly.
+	private static let conditionalActionIds: Set<String> = [
+		"download.current"
+	]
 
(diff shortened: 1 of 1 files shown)


skalthoff:

reviewer-session: wf_7dda2c5e-608-48
diff-stat: 1 file changed, 12 insertions(+), 3 deletions(-)

Checklist:
- [✓ N/A] error-swallowing — diff is test-only; no try?, empty catch, _ = try, or .ok() anywhere in the changed lines.
- [✓ N/A] MainActor-blocking FFI — diff is test-only; no core.X() calls added. The test class is @MainActor but AppModel() construction and paletteActions access are pure Swift computation, no FFI calls.
- [✓ N/A] paged-cache-only resolution — diff is test-only; no model.<things>.first { $0.id == X } added.
- [✓ N/A] optimistic-UI-without-echo — diff is test-only; no UI mutations.
- [✓ N/A] hotspot-file growth — only PaletteRowCoverageTests.swift touched; not a hotspot file.
- [✓ addressed] new-branch test coverage — the change is itself a test improvement. It unions conditionalActionIds (containing "download.current") into the asserted set so that removing "download.current" from actionTitleById/actionSymbolById while supportsDownloads=false would be caught. Preventive coverage, not reactive; no current test failure on main (the map entries were added by #942). The addition is justified.
- [✓ N/A] speculative scope — every changed line traces to the PR's explicit scope: (1) removing the stale #937 task-number reference from the class doc comment, and (2) adding conditionalActionIds to cover capability-gated actions. Nothing else touched.
- [✓ addressed] banned-comments — the PR removes the "exactly the `nav.favorites` regression in #937" task-number reference from the class doc comment (the former violation). The new doc comment on `conditionalActionIds` explains WHY the set exists (capability flags hide actions from the default roster), not WHAT the code mechanically does. No TODO or FIXME added.

Falsification: If a second capability-gated palette action is added (e.g., "export.playlist" behind supportsExportPlaylist), and its id is omitted from conditionalActionIds, the test would silently miss it when the flag is false. A tighter formulation would enumerate all known ids from actionTitleById.keys dynamically. However, this is an architectural improvement beyond scope — the static list is complete for the one capability-gated action that exists today ("download.current"). No invariant is violated by the current approach.

Outcome: approve

PR #1027, opened 2026-06-10
feat(dist): Sparkle delta updates in the release pipeline

### LINKED ISSUE #187: Delta updates for Sparkle (size optimization)
Sparkle supports binary delta updates (BSDiff over archive contents). A
Jellify `.dmg` is ~30–60 MB once Rust + Sparkle + fonts are bundled; deltas
typically shrink to <5 MB. `generate_appcast` will produce both full and
delta enclosures automatically if an adjacent older DMG is present in the
updates dir. Defer until 2–3 versions are out, then:

1. Keep the last N=4 DMGs in `./updates/`.
2. Re-run `generate_appcast` after each release.

No client changes needed — Sparkle picks the smallest path automatically.

---

<sub>Source: `05-macos-distribution.md`</sub>

skalthoff:
Sparkle supports binary delta updates that reduce update download size from ~30-60 MB (full DMG) to ~1-5 MB. This wires delta generation into the release pipeline so every release automatically produces and uploads deltas for the last 3 prior versions.

Closes #187

## What changed

**`macos/Scripts/generate-appcast.sh`**
- Removed the `--maximum-deltas 0` flag that had deliberately disabled delta generation (see comment referencing this issue: "Re-enable when the workflow learns to upload Sparkle's .delta artifacts")
- Set `--maximum-deltas "$DELTA_MAX"` (default 3) so `generate_appcast` generates `.delta` files into `$DMG_DIR` alongside the downloaded DMGs
- After `generate_appcast` runs, collect generated `.delta` files and write their paths to `$DELTA_MANIFEST` for the workflow to consume
- Add a second post-process URL injection pass for delta enclosure URLs (delta files don't exist as release assets yet at generation time, so the existing assets-loop can't inject their tag segment; this pass injects the latest tag for bare `.delta` enclosure URLs)
- Adjusted `gh release list --limit` from the hardcoded 10 to `DELTA_MAX + 4` so the fetch window scales with the configured delta depth
- Added `--dry-run` flag for local testing without credentials (`bash -n` + `shellcheck` clean)

**`.github/workflows/macos-release.yml`**
- Passes `DELTA_MANIFEST` env var to the `Generate appcast` step pointing to a temp file
- New `Upload Sparkle delta files to release` step runs after appcast generation: reads the manifest and uploads any `.delta` files to the current tag with `gh release upload --clobber`; no-op when manifest is absent or empty (first 1-3 releases)

## How deltas are bounded

`DELTA_MAX=3` (configurable via env) means `generate_appcast` produces at most 3 delta files per release (one for each of the 3 prior releases). Users upgrading from any of the last 3 versions get a ~1-5 MB delta; users on older versions get the full DMG. The `gh release list` fetch window is `DELTA_MAX + 4 = 7` releases.

## What was tested

- `bash -n` on `generate-appcast.sh`: pass
- `shellcheck` on `generate-appcast.sh`: pass
- `./macos/Scripts/generate-appcast.sh --dry-run`: pass (validates arg parsing and early exit)
- `python3 -c "import yaml,sys; yaml.safe_load(...)"` on `macos-release.yml`: pass

**What needs a live release to verify:**
- That `generate_appcast` actually writes `.delta` files into `$DMG_DIR` (requires 2+ real releases in the repo so the tool has prior DMGs to diff against)
- That delta enclosure URLs get the correct `v1.x.x` tag segment injected by the new delta post-process pass
- That `gh release upload` accepts the delta paths and they appear as downloadable assets on the release page
- That Sparkle clients on prior versions receive and apply a delta update (not the full DMG)

## Rollback

If delta generation causes appcast issues, set `DELTA_MAX=0` in the workflow env or revert the `--maximum-deltas` flag to 0 in `generate-appcast.sh`. No client-side changes are needed; Sparkle transparently falls back to the full DMG if no delta is available.


skalthoff:

reviewer-session: agent-ade4c7a174cec4ef1
diff-stat: 2 files changed, 123 insertions(+), 17 deletions(-) — .github/workflows/macos-release.yml (+37), macos/Scripts/generate-appcast.sh (+103/-17)

Checklist:
- [N/A] error-swallowing — shell/YAML only; no try?, .ok(), or swallowed errors. First-release no-op paths exit 0 cleanly.
- [N/A] MainActor-blocking FFI — no Swift code touched.
- [N/A] paged-cache-only resolution — no Swift code touched.
- [N/A] optimistic-UI-without-echo — no app state mutations.
- [N/A] hotspot-file growth — AppModel.swift, LyrebirdApp.swift, client.rs untouched.
- [N/A] new-branch test coverage — release pipeline change; no Rust/Swift test can exercise a CI workflow step. --dry-run covers the testable surface and was exercised. Live-release-only verification correctly disclosed.
- [N/A] speculative scope — every changed line is traceable to #187 (remove --maximum-deltas 0, DELTA_MANIFEST, upload step, delta URL injection pass).
- [FINDING] banned-comments — two issue-reference comments added by this PR, prohibited by CLAUDE.md ("Comments referencing the current task"):

  **macos/Scripts/generate-appcast.sh**, header block comment:
  ```
  #     the release workflow to upload. Closes #187.
  ```
  Fix: remove `. Closes #187` from that line.

  **macos/Scripts/generate-appcast.sh**, inline comment before the `generate_appcast` invocation:
  ```
  # include the correct tag segment (same as for the full DMG). Fixes #187.
  ```
  Fix: remove `. Fixes #187` from that line.

Falsification: on workflow re-run (workflow_dispatch on an existing tag where deltas were already uploaded to the release), the first URL-injection loop queries `gh release view` for existing assets and injects the tag into delta enclosure URLs; the new delta-specific injection pass then runs the same substitution. Because the sed pattern requires the filename to immediately follow `/releases/download/` (no slash in between), an already-tagged URL (`/releases/download/v1.x.x/Lyrebird...delta"`) won't match the pattern. The second pass is a safe no-op. This re-run path cannot be covered by a test but the invariant holds.

Outcome: request-changes — remove the two banned issue-reference comment suffixes, then re-request review.

Issue #1012, opened 2026-06-09
ArtistDetailView: About bio always shows Read more button even for short bios


skalthoff:
## Summary

`ArtistDetailView.aboutBody` unconditionally renders a \"Read more\" button whenever a non-empty `bioOverview` exists, even when the entire bio fits within the 4-line clamp. Users see a functional but misleading \"Read more\" button that opens a popover containing exactly the same text they already see in full — a confusing no-op affordance.

## Root cause

`ArtistDetailView.swift:868–888` (`aboutBody`):

```swift
// ArtistDetailView.swift:868
@ViewBuilder
private func aboutBody(overview: String, artistName: String) -> some View {
    VStack(alignment: .leading, spacing: 10) {
        Text(overview)
            .font(Theme.font(14, weight: .regular))
            .foregroundStyle(Theme.ink2)
            .lineLimit(4)
            .fixedSize(horizontal: false, vertical: true)
        Button {          // ← always rendered, no truncation check
            isBioExpanded = true
        } label: {
            Text("Read more")
                ...
        }
    }
}
```

The `AlbumDetailView.aboutBody` (lines 505–567) already has the correct pattern: it uses `AboutOverviewTruncation.isTruncated(clamped:full:)` (backed by two `GeometryReader` `PreferenceKey` measurements) and gates the button behind `if isOverviewTruncated { ... }`. That fix was tracked internally as `#475` in `AlbumDetailView.swift`. The artist bio never received the same treatment.

## Reproduction

Swift sequence:
1. Open the app, navigate to any artist with a short bio (≤ ~30 words, e.g. "1nonly" or "2 Mello" on `https://music.skalthoff.com`).
2. Scroll to the **About** section on `ArtistDetailView`.
3. Observe the **"Read more"** button below the bio text — the text fully fits in 4 lines.
4. Expected: button is absent (bio does not overflow). Observed: button always appears.

Server confirmation — both artists return short overviews via the real server:
```
curl "https://music.skalthoff.com/Artists?userId=<uid>&Fields=Overview&Limit=5&SortBy=SortName"
# → '1nonly': 31 words, '2 Mello': 33 words
```

## Fix sketch

Apply the same `AboutOverviewTruncation`/height-measurement pattern from `AlbumDetailView.swift:505–577` to `ArtistDetailView.aboutBody`:
- Add `@State private var isBioTruncated = false` (and the two height states).
- Wrap the button in `if isBioTruncated { ... }`.
[...]

PR #1029, opened 2026-06-10
feat(core): push playback status events over a callback stream

### LINKED ISSUE #433: Remove synchronous `core.status()` polling, use an event stream
`AppModel.startPolling` fires a 500-ms `Timer` that calls `core.status()` on the main queue, triggering `@Observable` downstream re-renders twice a second, even while the app is backgrounded, even while idle. The core status only *actually* changes on state transitions (play/pause/track end) plus a position tick from AVPlayer's own `addPeriodicTimeObserver` we already have.

Replace the poll with:
- A `uniffi::CallbackInterface PlayerObserver` the Rust core invokes on state transitions (already mutation points: `mark_state`, `set_current`, `clear`).
- Drop the `Timer` entirely. The position is already observed by AVPlayer's periodic observer; we can update `status.positionSeconds` there and coalesce into a published `@Observable` once per UI frame using `.animation(.linear(duration: 0.5))`.
- While the window is in the background (`NSApp.occlusionState` does not contain `.visible`), stop updating the UI-side `status` entirely.

Acceptance: idle CPU drops to ≤ 0.3% on an M1 in background, ≤ 1% in foreground with no playback. No wakeups from the `Timer`.

skalthoff:
The 1 Hz `core.status()` poll re-took the Rust player mutex and republished `@Observable` state every second — even paused, even in the background — so the core now pushes the changes instead and the timer is gone entirely.

Closes #433

## Event set

`PlayerObserver.player_changed(seq, kinds, status)` — registered via `set_player_observer` / `clear_player_observer`, carrying the full post-mutation `PlayerStatus` snapshot plus the changed facets:

- `StateChanged` — `mark_state`, `set_current` (forces Playing), `clear`/`stop`
- `TrackChanged` — identity changes from `set_current`, skips, `set_queue`, playhead priming on a cold queue, `clear`/`stop`
- `QueueChanged` — `set_queue`, `play_next`, `add_to_queue`, `clear_queue`, cursor moves from skips / in-queue `set_current`

Deliberately silent: `mark_position` (the platform's 1 Hz AVPlayer time observer is the position source — echoing it back would reintroduce the wake storm), and volume / shuffle / repeat / play-session-id setters (their UI call sites already refresh `status()` inline, and Control Center wants that round-trip synchronous). No-op mutations (re-marking the same state/track, empty queue inserts, repeat-one "skips") don't emit.

## Deadlock safety

No lock is ever held while a callback runs: each mutation computes `(seq, kinds, snapshot)` inside the player-lock scope and emits after release; the observer-slot lock is held only to clone the `Arc`, never across the call. `seq` is stamped *under* the player lock so it is monotonic in mutation order — the Swift side drops any push not strictly newer than the last applied, which makes the per-push main-actor hops safe against reordered delivery. A core test re-enters `status()`, `set_volume`, and `clear_player_observer` from inside the callback to pin the contract (it hangs rather than passes if emission ever moves back under a lock).

## Energy Impact

- Idle / paused: **zero** periodic work — the poll `Timer` no longer exists (rc13 baseline was ~3,600 wakes/h paused). No `Timer` remains anywhere in the status path.
- Playing: unchanged 1 Hz — the pre-existing, `rate == 0`-gated AVPlayer time observer now also forwards each tick (`AudioEngine.onPositionTick`, wired on both the AVQueuePlayer and DSP paths for reporting parity) to advance `status.positionSeconds`, the Dock ring, and the scrobble threshold without any FFI on the UI path.
- Events fire only on user/track actions, so the rc11 contract strictly improves; this also structurally fixes the rc12 frozen-PlayerBar failure mode since every `markState` transition publishes.
- Bonus correctness: seeks now mirror their target into `status` + `core.markPosition`, so a paused scrub no longer snaps back to the pre-seek position.

The synchronous `status()` FFI stays for inline refreshes (shuffle/repeat handlers, SmokeTest, seeding the surface at subscription start).

## Tests

- `core/src/tests/status_events.rs` (16 new): per-mutation emission kinds + snapshots, no-op silence, `mark_position`/volume/shuffle/repeat silence, seq monotonicity, observer replacement + unregistration, and the re-entrancy/deadlock pin.
- `macos/Tests/LyrebirdTests/PlayerEventStreamTests.swift` (7 new): end-to-end core-mutation→push→`status` through a real `AppModel`, stream teardown silence, stale-seq drops, track-change side effects (session history, Now Playing detail teardown), position-tick gating, paused-seek mirroring.
- Gates: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test` (332), `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`, `./macos/Scripts/build-core.sh`, clean `swift build` + `swift test` (1014) — all green, re-run after rebasing onto `main`.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -25,7 +25,9 @@ pub use enums::{ImageType, ItemField, ItemKind, ItemSortBy, SortOrder};
 pub use error::{LyrebirdError, Result};
 pub use library_cache::{LibrarySyncObserver, LibrarySyncSummary};
 pub use models::*;
-pub use player::{PlaybackState, Player, PlayerStatus, RepeatMode};
+pub use player::{
+    PlaybackState, Player, PlayerEventKind, PlayerObserver, PlayerStatus, RepeatMode,
+};
 pub use query::ItemsQuery;
 
 use crate::client::{JellyfinClient, PublicSystemInfo};
@@ -1584,6 +1586,31 @@ impl LyrebirdCore {
 
     pub fn status(&self) -> PlayerStatus {
         self.player.status()
+    }
+
+    /// Register `observer` to receive push notifications for player-state
+    /// changes (#433) — state transitions, current-track changes, and queue
+    /// changes — replacing the UI's 1 Hz `status()` poll loop. Replaces any
+    /// previously-registered observer.
+    ///
+    /// Contract highlights (full details on [`PlayerObserver`]):
+    /// * Callbacks fire synchronously on the mutating thread; marshal to
+    ///   your main thread and return quickly.
+    /// * No core lock is held during the callback, so re-entering any FFI
+    ///   from inside it is safe.
+    /// * `seq` is monotonic in mutation order — drop pushes whose `seq` is
+    ///   not greater than the last applied one.
+    /// * Position ticks never emit; the platform's own 1 Hz time observer is
+    ///   the position source. Volume / shuffle / repeat setters don't emit
+    ///   either — their UI call sites already refresh `status()` inline.
+    pub fn set_player_observer(&self, observer: Box<dyn PlayerObserver>) {
+        self.player.set_observer(Some(Arc::from(observer)));
+    }
+
+    /// Remove the observer registered via [`Self::set_player_observer`], if
+    /// any. Subsequent player-state changes emit nothing.
+    pub fn clear_player_observer(&self) {
+        self.player.set_observer(None);
     }
 
     // ======================================================================
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -7,6 +7,7 @@
 use crate::error::{LyrebirdError, Result};
 use crate::models::Track;
 use parking_lot::Mutex;
+use std::sync::Arc;
 
 #[derive(Clone, Debug, PartialEq, Eq, uniffi::Enum)]
 pub enum PlaybackState {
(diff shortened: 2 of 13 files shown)


skalthoff:

reviewer-session: agent-a967c9ddb14b18db3
diff-stat: +571 -137 Swift (AppModel+PlayerEvents new, AppModel/AppDelegate/Session/NowPlaying/PlaybackControls/AudioEngine modified) | +771 -90 Rust (player.rs, lib.rs, tests/status_events.rs) | +4 out-of-scope files

Checklist:

- [✓ N/A] error-swallowing — No `try?` swallowing real errors in new code. Two `try? await Task.sleep(...)` in test helper loops suppress CancellationError only, correct for deadline loops. No `.ok()` in Rust additions. N/A.

- [✓ N/A] MainActor-blocking FFI — `startStatusEventStream()` calls `core.setPlayerObserver` and `core.status()` on MainActor once at session start. `applyPlayerEvent` takes no FFI. `applySeekPosition` calls `core.markPosition` (trivial mutex-set, not a hot path). N/A.

- [✓ N/A] paged-cache-only resolution — No new `model.<things>.first { $0.id == X }` without fallback. N/A.

- [✓ N/A] optimistic-UI-without-echo — Status mutations flow core→observer→MainActor. The seek `applySeekPosition` writes both the reactive surface and `core.markPosition`. N/A.

- [✓ N/A] hotspot-file growth — `AppModel.swift` net +20 lines (necessary for the bridge + seq slot). `AudioEngine.swift` +15 for `onPositionTick`. `core/src/lib.rs` +28 for the two FFI entry points. All justified; `AppModel+NowPlaying.swift` loses 113 lines, offsetting growth elsewhere.

- [✓ N/A] new-branch test coverage — 16 Rust tests (`status_events.rs`) + 7 Swift tests (`PlayerEventStreamTests.swift`). Cover emission kinds, snapshots, no-op silence, seq monotonicity, re-entrancy/deadlock pin, stream lifecycle, stale-seq drop, track-change side effects, position-tick gating, paused-seek mirroring. All fail on `origin/main`. N/A.

- [✗ FINDING] speculative scope — Three groups of changes cannot be traced to issue #433 and are unacknowledged in the PR description:

  **1. `macos/Sources/Lyrebird/Screens/ArtistDetailView.swift` (−97 lines, +12):** Removes the `isBioTruncated`/`clampedBioHeight`/`fullBioHeight` height-measurement system added for issue #1012. Replaces the conditional "Read more" button (shown only when the bio genuinely overflowed 4 lines) with an unconditionally visible button. For artists with a short bio that fits in ≤4 lines, "Read more" now appears even though there is no text to expand. This is a functional regression on #1012 work.

  **2. `macos/Tests/LyrebirdTests/ArtistDetailTransportTests.swift` (−51 lines):** Removes four tests validating `AboutOverviewTruncation.isTruncated` — `testBioTruncatedWhenFullExceedsClamped`, `testBioNotTruncatedWhenHeightsMatch`, `testBioNotTruncatedBeforeMeasurementSettles`, `testBioEpsilonAbsorbsSubpointRounding`. These tests pass on `origin/main` and are unrelated to #433; deleting them removes coverage without replacement.

  **3. `macos/Scripts/generate-appcast.sh` (−103 lines) and `.github/workflows/macos-release.yml` (−37 lines):** Remove Sparkle binary delta generation and the "Upload Sparkle delta files to release" CI step. Unrelated to the player event stream.

- [✓ N/A] banned-comments — Issue number cross-references match project traceability convention. No what-code-does comments duplicating obvious logic. No new TODO/FIXME. N/A.

**Falsification (hypothetical failing test not covered):**

```swift
// A no-op mutation stamped seq=2 before a paused seek, but delivered after:
func testStalePositionInNewerSeqDoesNotOverwriteSeek() throws {
    let model = try AppModel()
    model.applyPlayerEvent(seq: 1, kinds: [.trackChanged, .stateChanged],
        status: makeStatus(state: .paused, track: makeTrack("a"), position: 10))
    model.seek(toSeconds: 90)
    // seq=2 push carries old positionSeconds=10 (was stamped before the seek
    // but delivered after). The seq guard (seq > 1) passes, so the push
    // applies, overwriting 90 → 10.
    model.applyPlayerEvent(seq: 2, kinds: [.stateChanged],
        status: makeStatus(state: .paused, track: makeTrack("a"), position: 10))
    XCTAssertEqual(model.status.positionSeconds, 90)  // fails: returns 10
}
```
The seq-drop guard doesn't protect against a legitimately newer push that carries a stale position snapshot from before the seek. This is a real but low-frequency race. Not the blocking finding — noted for awareness.

**Outcome: request-changes**

Required before merge: remove `ArtistDetailView.swift`, `ArtistDetailTransportTests.swift`, `generate-appcast.sh`, and `macos-release.yml` from this PR's diff. The #1012 bio-truncation removal is a functional regression; it belongs in a dedicated PR if intentional. The release infrastructure changes belong separately.

Issue #239, opened 2026-04-22
Collaborative / public indicators


skalthoff:
- If Jellyfin marks the playlist as shared, show a `users` icon + "Shared with 3 people" under the title.
- Owner badge next to contributed tracks.

---

## 7. Search

**Brief.** Heavy lifter. 18pt-type search field in a 14pt-radius pill. Instant-as-you-type (250ms debounce, 2-char min). Results group into Top Result (hero card), Artists, Albums, Tracks, Playlists, Genres — each an independent `Section` with "See all" link. Scope chips at top (All / Artists / Albums / Tracks / Playlists / Podcasts). Recent searches as history chips when the field is empty.

<sub>Source: `06-screen-specs.md`</sub>

PR #1036, opened 2026-06-10
feat(macos): graceful shutdown — stop playback, persist queue, final report

### LINKED ISSUE #446: Graceful shutdown — flush DB, stop AVPlayer, persist queue
On `applicationShouldTerminate` / `NSApp.terminate`, we currently yank AVPlayer out from under its observer, leak the keyring connection, and don't persist the user's queue.

- Register `NotificationCenter.default.addObserver(forName: NSApplication.willTerminateNotification)` that: calls `audio.stop()`, calls `core.persist_queue()` (new Rust entry point — writes the current queue snapshot to `settings` as JSON), `PRAGMA optimize` + `wal_checkpoint(TRUNCATE)` on the SQLite connection.
- On next launch, if a persisted queue exists: restore silently but do NOT auto-play. Show a toast "Resume where you left off?" with a play button.
- Handle force-quit-while-downloading (when we add downloads): the download worker writes to a `.part` file + `download_state` table row; on restart, any row with `state IN ('downloading', 'pending')` resumes via HTTP `Range`.

Acceptance: `kill -9` mid-playback followed by a fresh launch does not corrupt the DB (WAL recovery is transparent), restores queue state, and allows one-click resume.

skalthoff:
Quitting mid-playback dropped the final position mark and server stop report — resume points were stale and Now Playing lingered on other clients.

Closes #446

- `applicationShouldTerminate` → bounded shutdown sequence (hard ~2s timeout, never hangs quit): stop playback (both AVQueuePlayer and DSP paths), final `reportPlaybackStopped` + position mark, queue persistence
- `AppModel+Shutdown.swift` extension; testable sequence helper

Gates: clean swift build; swift test 1087/1087 (verified post-stage).

### DIFF
--- a/macos/Sources/Lyrebird/AppDelegate.swift
+++ b/macos/Sources/Lyrebird/AppDelegate.swift
@@ -176,10 +176,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         }
 
         // Step 2 (synchronous, main actor): capture queue state to
-        // UserDefaults. Must run before the player is torn down further
-        // (the position is still readable at this point because `stop()`
-        // clears the player items but `status.positionSeconds` is a core
-        // mirror that `core.stop()` zeroes after our call).
+        // UserDefaults. `status.positionSeconds` still holds the last
+        // position tick — player-event delivery re-enters the main actor
+        // asynchronously and cannot land while this method occupies it.
         if steps.contains(.persistQueue) {
             model.persistQueueSnapshot()
         }
--- a/macos/Sources/Lyrebird/AppDelegate.swift
+++ b/macos/Sources/Lyrebird/AppDelegate.swift
@@ -70,6 +70,19 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// a DHCP / Wi-Fi link before we attempt TCP to the Jellyfin endpoint.
     private let wakeDebounceNanos: UInt64 = 2_000_000_000
 
+    /// Hard deadline for the async shutdown sequence. If `audio.stop()` /
+    /// `POST /Sessions/Playing/Stopped` has not completed within this window
+    /// we reply `.terminateNow` and let the OS finish the job. 2 s matches
+    /// the upper end of a server round-trip on a LAN while keeping the quit
+    /// snappy from the user's perspective.
+    private let shutdownTimeoutNanos: UInt64 = 2_000_000_000
+
+    /// `true` once `applicationShouldTerminate` has dispatched the async
+    /// shutdown work. Guards against a second call (which AppKit can make if
+    /// the app calls `NSApp.terminate` from a menu action while a prior quit
+    /// is already in flight).
+    private var shutdownInFlight: Bool = false
+
     // MARK: - NSApplicationDelegate
 
     func applicationDidFinishLaunching(_ notification: Notification) {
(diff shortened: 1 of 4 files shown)


skalthoff:

diff-stat: 18 files changed, 848 insertions, 435 deletions

Checklist:

- [✓ N/A] error-swallowing — `try? await Task.sleep(...)` appears twice (AppDelegate.swift:192, AppModel+Shutdown.swift:304). `Task.sleep` only throws `CancellationError`; swallowing cancellation in a shutdown timer and a post-stop seek delay is intentional and harmless.

- [✓ N/A] MainActor-blocking FFI — `model.stop()` and `model.persistQueueSnapshot()` are called on the main actor inside `applicationShouldTerminate`, but neither makes a live FFI call on the calling thread. `audio.stop()` dispatches `reportPlaybackStopped` as `Task.detached`, and `persistQueueSnapshot` writes to `UserDefaults` (no Rust mutex). No beach-ball risk.

- [✓ N/A] paged-cache-only resolution — the restore path (`applyQueueSnapshot`) hydrates `upNextUserAdded`/`upNextAutoQueue` from the persisted snapshot directly; it does not resolve IDs through the live cache. The `resumeFromSnapshot` path calls `play(tracks:)` with the materialized `Track` objects. No cache-miss failure mode.

- [✓ N/A] optimistic-UI-without-echo — `persistQueueSnapshot` writes to `UserDefaults` synchronously with no server echo needed. The `applyQueueSnapshot` path does not mutate server state. No mutation without server echo.

- [✓ N/A] hotspot-file growth — `AppModel.swift` net +3 lines (adds `pendingResumeSnapshot`, removes `showingNewPlaylistSheet`). New behavior lives in `AppModel+Shutdown.swift`. Justified.

- [✓ addressed] new-branch test coverage — `GracefulShutdownTests.swift` (403 lines, 22 tests) added; covers `shouldPersistSnapshot`, `shutdownSteps` ordering, `PersistedTrack` Codable round-trip, `QueueSnapshot` Codable, persist/load round-trip, idle-clears-stale, malformed-JSON auto-clears, `applyQueueSnapshot` hydration, `dismissResumeSnapshot`, `resumeFromSnapshot`. Tests are real (no mocked AppModel — constructs live instances with temp data dir).

- [✗ finding] speculative scope — **The diff touches 11 files with no connection to issue #446.** The `is_public`/`set_playlist_visibility` removal spans:
  - `core/src/models.rs` — removes `Playlist.is_public` field from a `uniffi::Record`
  - `core/src/lib.rs` — removes `set_playlist_visibility` FFI function and `is_public` param from `create_playlist`
  - `core/src/client.rs` — removes `set_playlist_visibility` async fn and `CreatePlaylistBody.is_public`
  - `core/src/tests/playlists.rs` — removes 160 lines of visibility tests
  - `macos/Sources/Lyrebird/AppModel+Playlists.swift` — removes `setPlaylistVisibility`, reworks `createPlaylist`/`duplicatePlaylist`/`renamePlaylist`/`beginNewPlaylist`
  - `macos/Sources/Lyrebird/AppModel+PlaylistTracks.swift` — drops `isPublic:` from all `Playlist()` constructors
  - `macos/Sources/Lyrebird/AppModel+Queue.swift` — drops `isPublic:` from `saveQueueAsPlaylist`
  - `macos/Sources/Lyrebird/AppModel+Resolvers.swift` — drops `isPublic:` from `Playlist()` in resolver
  - `macos/Sources/Lyrebird/Components/PlaylistContextMenu.swift` — removes Make Public/Private toggle
  - `macos/Sources/Lyrebird/Screens/MainShell.swift` — removes the `NewPlaylistSheet` sheet
  - `macos/Sources/Lyrebird/Screens/NewPlaylistSheet.swift` — file deleted
  - `macos/Sources/Lyrebird/Components/Sidebar.swift` — reworks `beginNewPlaylist` to inline edit

  None of this is named in #446. The PR body claims only `Closes #446`. This is not a necessary prerequisite for graceful shutdown; it is a separate feature removal bundled into the same commit. Per the scope rule, every changed line must be traceable to the linked issue.

- [✗ finding] banned-comments — `AppDelegate.swift` lines 180-182 contain a factually wrong explanatory comment: "the position is still readable at this point because `stop()` clears the player items but `status.positionSeconds` is a core mirror that `core.stop()` zeroes **after our call**". `core.stop()` is called synchronously inside `audio.stop()` (AudioEngine.swift:782), which is part of Step 1. By the time Step 2 reads `status.positionSeconds`, `core.stop()` has already run and zeroed the Rust-side `position_seconds`. The position is valid on the Swift side because `applyPlayerEvent` arrives via an async `Task` hop that hasn't fired yet — not because of the claimed ordering. The comment misstates the causal mechanism.

---

Falsification — hypothetical failing test this PR doesn't cover:

```swift
func testPersistCapturesNonZeroPositionAfterStop() throws {
    let model = try AppModel()
    // Set the Swift-side mirror at a non-zero position.
    model.status.currentTrack = makeTrack(id: "p1")
    model.status.positionSeconds = 42.5

    // After stop() + persistQueueSnapshot(), positionSeconds must have been
    // captured before the async applyPlayerEvent Task zeroed the mirror.
    model.persistQueueSnapshot()

    let loaded = AppModel.loadQueueSnapshot()
    XCTAssertEqual(loaded?.positionSeconds, 42.5,
        "position must be captured at the value set before stop was called")
}
```

This test is absent. The existing `testPersistAndLoadRoundTrip` never sets `status.positionSeconds` and only checks track IDs. No test verifies that the position value > 0 survives the stop→persist ordering.

---

Outcome: request-changes

Required:
1. Split the playlist-visibility removal into a separate PR linked to its own issue (e.g. #239 or a new one), or remove those changes entirely from this branch so it contains only #446 work.
2. Fix `AppDelegate.swift` lines 180-182: the comment must correctly state that `status.positionSeconds` is a Swift-side mirror set by the 1Hz `handlePositionTick`, and it remains at its last tick value because the `applyPlayerEvent` Task hop is enqueued but not yet delivered when `persistQueueSnapshot` runs synchronously on the main actor — not because `core.stop()` zeroes happen after Step 2.

Issue #335, opened 2026-04-22
Visible, themed focus ring


skalthoff:
The default macOS focus ring is system blue, which clashes with the deep
purple theme. Either accept the system ring (fine, Apple prefers this), or
define a custom ring via `.focusEffect(.automatic)` with a rectangle overlay
driven by `@FocusState`. Whatever we pick must:

1. Be at least 2 px thick, offset 2 px from the control.
2. Maintain ≥ 3:1 contrast against `Theme.bgAlt` (#140B30) — the existing
   `Theme.accentHot` (#FF066F) passes at ≈ 7.8:1.
3. Never be removed wholesale with `.focusEffectDisabled()` on a focusable
   control.

**Acceptance criteria**
- All focusable controls show a visible focus ring when focused via Tab.
- Ring color validated with Colour Contrast Analyser against the surface
  behind it.

---

<sub>Source: `07-a11y-i18n.md`</sub>

Issue #318, opened 2026-04-22
Sidebar auto-hide + resize


skalthoff:
- Drag the sidebar's right edge to resize 200–320pt.
- Toggleable "Auto-hide" in Preferences.

<sub>Source: `06-screen-specs.md`</sub>

PR #1051, opened 2026-06-10
polish(macos): declutter main shell — dead focus rings, duplicate toggle, breadcrumbs, inspector duplication

### LINKED ISSUE #1050: Main shell visual clutter: dead region-focus rings, duplicate sidebar toggle, redundant breadcrumb strip, queue duplicated beside Now Playing / Play Queue
UX audit of the main shell found four compounding sources of visual clutter and distracting focus behavior:

1. **Whole-pane focus rings + dead ⌃Tab region cycle.** `MainShell` marks the sidebar, content pane, queue inspector, and player bar `.focusable()` and draws a custom 2pt `RegionFocusRing` around the focused pane, without `.focusEffectDisabled()` — so a focused pane can show the system ring *and* the brand ring around the entire region. The ⌃Tab cycle that drives it is inert in the default config: `.onKeyPress` only fires once something inside the shell has focus, and the toolbar search field is AppKit-hosted, so the hop never triggers for mouse-first users (verified live: ⌃Tab does nothing with and without the field focused). Net effect: no working feature, but whole panes can light up with double rings when focus lands on their containers. `themedFocusRing()` (#335) has zero adopters.

2. **Duplicate sidebar-toggle buttons.** `NavigationSplitView` provides the standard toggle; `MainShell` adds a second custom `ToolbarItem(placement: .navigation)` with the same `sidebar.left` glyph. Both render side by side (verified live).

3. **Breadcrumb strip is redundant where shown, hidden where useful.** The `topBar` breadcrumbs render only at NavigationStack root — i.e. "Lyrebird > Library" while the sidebar already highlights Library and the page already shows a "Library" title + "YOUR LIBRARY" eyebrow. On drill (album/artist/playlist), where a trail would help, pushed destinations cover the bar entirely, so the name-resolution half of `breadcrumbSegments` (and `resolvedNameCache` feeding it) is dead code in practice.

4. **Queue inspector duplicates whole surfaces.** The 320pt `QueueInspector` stays mounted while `Route.fullQueue` (a full-page queue) or `Route.nowPlaying` (which embeds a Queue tab) is pushed — queue-beside-queue, and now-playing artwork/meta shown 2–3× across one window (page + inspector + player bar).

Fix: remove the region-cycle apparatus, pane `.focusable()` wrappers and both unused focus-ring modifiers; drop the custom toolbar toggle; remove the breadcrumb strip and its dead resolvers; suppress the inspector while the two queue-bearing routes are on top (persisted toggle untouched).

skalthoff:
Closes #1050.

UX-audit cleanup of the main shell's clutter and focus behavior. Four changes, all verified against the live app before/while editing:

**1. Remove the ⌃Tab region cycle + whole-pane focus rings (−FocusRegion.swift, −RegionFocusRing/ThemedFocusRing).**
The cycle is unreachable in the default config: `.onKeyPress(.tab)` only sees events once something inside the SwiftUI hierarchy has focus, and the toolbar search field is NSToolbar-hosted, so pressing ⌃Tab with or without the field focused does nothing (verified live). What the apparatus *did* do is make all four shell panes `.focusable()` and draw a 2pt brand ring around whichever pane container picked up focus — without `.focusEffectDisabled()`, so the system ring could draw on top of the same pane. `themedFocusRing()` (#335) had zero call sites. Switch Control groupings, accessibility labels, and sort priorities are untouched; standard control-level focus traversal is unaffected.

**2. Drop the custom sidebar-toggle toolbar button.**
`NavigationSplitView` already renders the standard toggle, so the shell showed two identical `sidebar.left` buttons side by side. ⌘⌥S / View ▸ Show Sidebar still routes through `toggleSidebarManually()` and keeps recording the auto-hide manual-override flag.

**3. Remove the breadcrumb strip (−Breadcrumbs.swift, −breadcrumb name resolvers, −resolvedNameCache).**
The bar only rendered at NavigationStack root, where it read "Lyrebird > Library" under a toolbar, over a "Library" page title, beside a sidebar already highlighting Library. On drill — the only place a trail would help — pushed destinations covered it entirely, so the album/artist segment-name resolution chain (and the write-only `resolvedNameCache` feeding it) was dead in practice.

**4. Hide the Queue Inspector while Now Playing / Play Queue routes are on top.**
The 320pt panel stayed mounted beside `Route.fullQueue` (queue-page-beside-queue-pane) and `Route.nowPlaying` (which embeds its own Queue tab), so one window showed the same queue/artwork two to three times. The panel now yields while those routes are pushed; the persisted `isQueueInspectorOpen` choice is untouched and the panel returns on pop.

Net −723 lines. `swift build` clean; all 1112 package tests pass.


skalthoff (commented):

reviewer-session: sonnet-4-6-adversarial
diff-stat: 10 files changed, 30 insertions(+), 753 deletions(−)

Checklist:

- [✓ N/A] error-swallowing — No new `try?`, empty `catch`, or `_ = try` patterns introduced. The `resolveAlbum` simplification (removing the `resolvedNameCache` write) keeps the existing `catch { _ = handleAuthError(error); return nil }` path intact.

- [✓ N/A] MainActor-blocking FFI — No new `try core.X(...)` calls added. The only new code is the 6-line `isInspectorVisible` computed property on `MainShell`, which reads only `model.isQueueInspectorOpen` (a Bool) and `model.navPath.last` (an in-memory array) — no FFI.

- [✓ N/A] paged-cache-only resolution — `resolvedNameCache` is removed, but `resolveAlbum` and `resolveArtist` still fall back to a core FFI call (`core.fetchItem`) on cache miss. The deleted cache was only a secondary seeded write layered on top of those resolvers; the primary fallback path (cache-first → FFI on miss) is intact in `AppModel+Resolvers.swift`.

- [✓ N/A] optimistic-UI-without-echo — No new mutations added by this PR. The `isInspectorVisible` guard is read-only display logic.

- [✓ N/A] hotspot-file growth — `AppModel.swift` loses 7 lines (net shrink), `MainShell.swift` loses 259 lines (net shrink). `client.rs` is not touched.

- [✓ N/A] new-branch test coverage — The two deleted test files (`FocusRegionCycleTests.swift`, `BreadcrumbNameResolutionTests.swift`) covered deleted code and are correctly removed. The new `isInspectorVisible` property is a private computed property on a SwiftUI `View`; its logic is a 2-branch enum pattern match over a typed `[Route]` array. The property cannot be unit-tested without extracting it from the view layer; this is consistent with how other view-display predicates in this codebase are treated. The behavioral properties it reads (`isQueueInspectorOpen`, `navPath`) are tested in `ViewMenuBindingsTests.swift`.

- [✓ N/A] speculative scope — Every changed line traces to one of the four stated cleanup items: (a) region-cycle + focus ring removal, (b) duplicate sidebar toggle removal, (c) breadcrumb strip + `resolvedNameCache` removal, (d) queue inspector suppression on queue routes. No unrelated lines found.

- [✓ N/A] banned-comments — No new comments explaining *what* code does mechanically, no task-references added fresh (`(#318)` in the `toggleSidebarManually` doc-comment is a pre-existing cross-ref, not newly introduced), no new TODO/FIXME. The comment at `AppModel+Resolvers.swift:259` ("so breadcrumbs can read the name") is pre-existing on `origin/main` and was not touched by this PR.

Orphaned-symbol audit (all confirmed clean):
- `FocusRegion`, `FocusRegionCycle`: zero remaining references
- `RegionFocusRing`, `ThemedFocusRing`, `themedFocusRing`, `regionFocusRing`: zero remaining references
- `Breadcrumbs` (UI struct), `breadcrumbAlbumName`, `breadcrumbArtistName`, `resolvedNameCache`: zero remaining references
- `Theme.focusRing`, `focusRingHighContrast`: zero remaining references
- `topBar` (removed view builder): zero remaining references
- `cycleRegion`, `handleTabKey`, `currentRegion`: zero remaining references

`isInspectorVisible` correctness: `navPath` is `[Route]` (non-optional), so `.last` returns `Route?`. Pattern matching `case .nowPlaying?, .fullQueue?:` is correct Swift optional-unwrapping syntax for a no-associated-value enum case. `animation(value:)` updated consistently from `model.isQueueInspectorOpen` to `isInspectorVisible`.

`toggleSidebarManually()` still live: call site at `MainShell.swift:242` (via `.onChange(of: model.sidebarToggleRequest)`), definition at line 372, menu path wired in `LyrebirdApp.swift:436`. Not orphaned.

Accessibility: sidebar `.accessibilityElement(children: .contain)` + `.accessibilityLabel` + `.accessibilitySortPriority(100)` intact at lines 82-89. Content column `.accessibilityElement(children: .contain)` + `.accessibilityLabel` + `.accessibilitySortPriority(90)` intact at lines 151-156. PlayerBar sort priority at line 199. Switch Control groupings are unaffected.

Falsification: no such case exists for `isInspectorVisible` as a build-breakable invariant. The property is a pure read of two typed model properties over a closed enum. A hypothetical test would check `isQueueInspectorOpen=true, navPath=[.artist("x")]` → visible; `navPath=[.nowPlaying]` → suppressed; `navPath=[.fullQueue]` → suppressed. All three branches are covered by the implementation's exhaustive `switch`. The enum is `sealed` (no `@unknown default` needed since it's not `@objc`), so new routes added later that don't match `.nowPlaying`/`.fullQueue` correctly fall through to `default: return true`.

Outcome: approve

PR #1056, opened 2026-06-10
fix(macos): drop the launch splash once the session restores instead of after the first library fetch

### LINKED ISSUE #1055: Launch splash stays up through the full first library fetch (~20s on large libraries)
From the UX audit (live observation against a ~20k-album library): cold launch showed the blank "Lyrebird" brand splash for ~20 seconds before any UI appeared.

Cause: `attemptRestoreSession` (AppModel+Session.swift) only clears `isRestoringSession` via `defer`, i.e. after `refreshLibrary()` and `refreshDownloads()` complete — so the splash covers the whole first library round-trip, not just the session decision. The shell already has proper progressive-loading UX for that phase (library skeletons, cache-first paint from the local DB cache), so the splash buys nothing and reads as a hang.

Fix: flip `isRestoringSession = false` immediately after the restored session is assigned (before the library fetch). The `defer` keeps covering the no-session and error exits.

skalthoff:
Closes #1055. Third PR from the UX audit (#1050, #1052).

Cold launch on a large library sat on the blank "Lyrebird" splash for ~20s (observed live against a ~20k-album server) because `attemptRestoreSession` only cleared `isRestoringSession` in its `defer` — after `refreshLibrary()` + `refreshDownloads()` finished. The splash now drops as soon as the restored session is assigned; the shell takes over with its existing progressive-loading UX (library skeletons + cache-first paint from the local DB cache). The `defer` still clears the flag on the no-session and error paths, and the duplicate `false` write is idempotent.

One-hunk change. `swift build` clean; all 1119 package tests pass.

### DIFF
--- a/macos/Sources/Lyrebird/AppModel+Session.swift
+++ b/macos/Sources/Lyrebird/AppModel+Session.swift
@@ -163,14 +163,9 @@ extension AppModel {
             self.serverURL = session.server.url
             self.username = session.user.name
             self.errorMessage = nil
-            // Drop the launch splash as soon as the session is decided —
-            // the shell renders its own loading skeletons (and paints from
-            // the local library cache) while the fetches below run. Holding
-            // the flag through `refreshLibrary()` kept a blank brand splash
-            // up for the whole first library round-trip, which on a large
-            // library reads as a hang. The `defer` above still clears the
-            // flag on every other exit path; re-assigning `false` is
-            // idempotent.
+            // The splash must end at the session decision, not after the
+            // library fetch — the shell owns that phase with skeletons and
+            // the cache-first paint.
             isRestoringSession = false
             startStatusEventStream()
             await refreshLibrary()
--- a/macos/Sources/Lyrebird/AppModel+Session.swift
+++ b/macos/Sources/Lyrebird/AppModel+Session.swift
@@ -163,6 +163,15 @@ extension AppModel {
             self.serverURL = session.server.url
             self.username = session.user.name
             self.errorMessage = nil
+            // Drop the launch splash as soon as the session is decided —
+            // the shell renders its own loading skeletons (and paints from
+            // the local library cache) while the fetches below run. Holding
+            // the flag through `refreshLibrary()` kept a blank brand splash
+            // up for the whole first library round-trip, which on a large
+            // library reads as a hang. The `defer` above still clears the
+            // flag on every other exit path; re-assigning `false` is
+            // idempotent.
+            isRestoringSession = false
             startStatusEventStream()
             await refreshLibrary()
             await refreshDownloads()
--- a/macos/Tests/LyrebirdTests/SessionRestoreTests.swift
+++ b/macos/Tests/LyrebirdTests/SessionRestoreTests.swift
@@ -1,0 +1,44 @@
+import XCTest
+
+@testable import Lyrebird
+@testable import LyrebirdCore
+
+/// Launch-splash lifecycle around `attemptRestoreSession`.
+///
+/// The flag starts `true` (so the first render is the splash, never a
+/// `LoginView` flash) and must be `false` after the restore pass on every
+/// exit path. The no-persisted-session path is driven for real against a
+/// throwaway data dir; the restored-session happy path has no core
+/// injection seam (`AppModel.init` constructs its own `LyrebirdCore`), so
+/// its early-flip ordering — splash down before the first library fetch —
+/// is verified manually, not here.
+@MainActor
+final class SessionRestoreTests: XCTestCase {
+
+    override class func setUp() {
+        super.setUp()
+        let dir = NSTemporaryDirectory() + "lyrebird-tests-\(UUID().uuidString)"
+        setenv("XDG_DATA_HOME", dir, 1)
+    }
+
+    func testSplashFlagStartsRaised() throws {
+        let model = try AppModel()
+        XCTAssertTrue(model.isRestoringSession)
+    }
+
+    func testRestoreWithNoPersistedSessionLowersSplashFlag() async throws {
+        let model = try AppModel()
+        await model.attemptRestoreSession()
+        XCTAssertFalse(model.isRestoringSession)
+        XCTAssertNil(model.session)
+    }
+
+    func testSecondRestoreAttemptIsNoOp() async throws {
+        let model = try AppModel()
+        await model.attemptRestoreSession()
+        // A re-fired `.task` must not re-raise the splash or re-run restore.
+        model.isRestoringSession = false
+        await model.attemptRestoreSession()
+        XCTAssertFalse(model.isRestoringSession)
+    }
+}


skalthoff (commented):

reviewer-session: sonnet-4-6
diff-stat: 1 file changed, 9 insertions(+) — AppModel+Session.swift only

---

### Pre-flight verification (per instructions)

**1. RootView branching / LoginView flash** — CLEAN.
In the new ordering: `self.session = session` (line 162) is assigned BEFORE `isRestoringSession = false` (line 174). When the flag flips on the main actor, `session != nil` is already true. The `RootView.body` if-chain checks `isRestoringSession` first; when that becomes false, it falls through to the `session == nil` branch — which is already false. MainShell mounts; LoginView is never reachable in this path.

**2. defer coverage** — CLEAN.
`defer { isRestoringSession = false }` at line 156 is in the function body scope, not inside the `do` block, so it fires on all exit paths: the catch path, the `guard let session = restored else { return }` nil path, and the happy path (where the explicit assignment on line 174 fires first and the defer is idempotent). The `guard session == nil` early return at lines 152-154 sets the flag explicitly before defer is registered — that's safe and intentional (avoids the flag staying true if a session is already live).

**3. refreshLibrary/refreshDownloads MainActor-safety** — CLEAN.
Both functions are methods on `@MainActor final class AppModel`. They are already called while a session is live and MainShell is mounted from: `AppModel+Session.swift` line 42-43 (login path), line 25 (network retry), and `OnboardingView.swift` lines 647 and 655. The pattern of calling these functions mid-session is well-established.

**4. Test coverage** — REAL FINDING (see below).

**5. MainActor isolation** — CLEAN.
`AppModel` is `@MainActor final class`; `attemptRestoreSession` carries no `nonisolated` annotation. Extensions of a `@MainActor` type inherit main-actor isolation in Swift concurrency. `isRestoringSession = false` at line 174 therefore runs on the main thread, and the `@Observable` write posts correctly for the `.animation(.default, value: model.isRestoringSession)` modifier in LyrebirdApp.swift line 874.

---

### Checklist

- [✓ N/A] **error-swallowing** — No `try?` or empty catch added by this diff. The two existing `try? core.logout()` / `try? core.forgetToken()` are pre-existing and not touched.

- [✓ N/A] **MainActor-blocking FFI** — The diff adds no FFI call. `isRestoringSession = false` is a pure property write.

- [✓ N/A] **paged-cache-only resolution** — No resolver logic touched.

- [✓ N/A] **optimistic-UI-without-echo** — No local-state mutation without a server round-trip. The session was already written by the core FFI; this is only a UI flag flip.

- [✓ N/A] **hotspot-file growth** — The modified file is `AppModel+Session.swift`, not the hotspot trio (`AppModel.swift`, `LyrebirdApp.swift`, `client.rs`). +9 lines in an extension file.

- [✗ FINDING] **new-branch test coverage** — The PR changes observable behavior: `isRestoringSession` now transitions to `false` before `refreshLibrary()` completes, not after. Zero tests in `macos/Tests/LyrebirdTests/` reference `isRestoringSession`, `attemptRestoreSession`, or `RestoreLoadingView` on either branch. The fixer has not stated "this is exclusively manual"; no test was added. A Swift test that constructs a mock AppModel, calls `attemptRestoreSession`, and asserts (a) the flag is false and session is non-nil at the point refreshLibrary is entered, and (b) the flag stays false on the catch path would be appropriate here.

- [✓ N/A] **speculative scope** — Every added line is traceable to the stated problem: the launch splash holding past the first library round-trip. No unrelated behavior changed.

- [✗ FINDING] **banned-comments** — Lines 166-173 of `AppModel+Session.swift` (the added block) contain two banned sentences:
  - "The `defer` above still clears the flag on every other exit path; re-assigning `false` is idempotent." — This explains *what the code does*. The code already shows that; the comment is redundant prose.
  - "Holding the flag through `refreshLibrary()` kept a blank brand splash up for the whole first library round-trip, which on a large library reads as a hang." — This describes the prior behavior in task-context language ("reads as a hang") that references the current fix. CLAUDE.md prohibits comments that reference the current task or explain what code does.
  The entire block can be trimmed to nothing, or at most a one-liner: `// Early dismiss — shell paints from cache while fetches run.`

---

### Falsification sketch

A test this PR does not cover:

```swift
// Hypothetical: what if resumeSession() throws after a partial write?
// Place in macos/Tests/LyrebirdTests/SessionRestoreTests.swift
func testIsRestoringSessionFalseOnResumeThrow() async throws {
    // Arrange: core stub that throws on resumeSession
    let stub = CoreStub(resumeSessionResult: .failure(LyrebirdError.Network))
    let model = AppModel(core: stub)
    XCTAssertTrue(model.isRestoringSession)

    // Act
    await model.attemptRestoreSession()

    // Assert: flag must be false — defer should have cleared it
    XCTAssertFalse(model.isRestoringSession)
    XCTAssertNil(model.session)
}
```

This test passes before AND after the PR (defer handles the throw path correctly), but it doesn't exist. The actual gap is the *success* path: no test asserts that the flag is false and `session != nil` simultaneously at the moment refreshLibrary begins — which is the new invariant this PR introduces. A cancellation of the `.task` midway through `refreshLibrary` (e.g., the view disappears) could leave `session` set but `isRestoringSession` already false; the defer fires on task cancellation and is again idempotent — but this is never verified.

---

### Outcome

**request-changes**

Two findings:
1. **Missing test** (checklist item 6): No test validates the new early-flag-flip behavior. State "manual only" or add a test.
2. **Banned comments** (checklist item 8): Lines 166-173 contain explanation of what the code does and a task-context reference. Per CLAUDE.md these must be removed or stripped to a minimal why-line.

skalthoff:
Addressed both findings:
- Comment trimmed to the constraint only (why the flip precedes the fetch), self-justifying lines dropped.
- Added `SessionRestoreTests` (3 tests): flag starts raised, the no-persisted-session restore path lowers it (pins the `defer`), and a re-fired restore attempt is a no-op. The restored-session happy path has no core-injection seam (`AppModel.init` constructs its own `LyrebirdCore`), so the early-flip ordering there is declared manual-only — verified live against the test server — and is documented as such in the test header.

skalthoff:
## Adversarial review — refix delta verification (Sonnet follow-up)

reviewer-session: sonnet-4-6
delta-stat: AppModel+Session.swift (comment block reduced), SessionRestoreTests.swift (44 lines, new file)

---

### Finding 1 re-check: banned-comment

The 8-line block is now 3 lines (AppModel+Session.swift lines 166-168):

```
// The splash must end at the session decision, not after the
// library fetch — the shell owns that phase with skeletons and
// the cache-first paint.
```

This states *why* the ordering is what it is (design intent / invariant), not *what the code does*. The sentences that explained the `defer` behavior and referenced the task context ("reads as a hang") are gone. No banned-comment violation remains.

### Finding 2 re-check: test coverage

Three tests added in `SessionRestoreTests.swift`:

- `testSplashFlagStartsRaised` — confirms `isRestoringSession` initialises `true`. Pins the CLAUDE.md gap #1 invariant (first render is never a LoginView flash).
- `testRestoreWithNoPersistedSessionLowersSplashFlag` — drives `attemptRestoreSession()` against a real core instance with a fresh throwaway data dir. Asserts flag is `false` and `session` is nil after the call. This is the defer path (resumeSession returns nil → `guard let session = restored else { return }` → defer fires). Directly covers the no-session exit.
- `testSecondRestoreAttemptIsNoOp` — confirms `hasAttemptedRestore` blocks re-entry; the flag cannot be re-raised by a re-fired `.task`.

**Manual-only declaration for the happy path**: the doc comment on the test class states explicitly: "its early-flip ordering — splash down before the first library fetch — is verified manually, not here" with the stated reason ("AppModel.init constructs its own LyrebirdCore"). This satisfies the "add a test or declare manual-only explicitly" requirement from the previous review. The declaration is accurate — the XDG isolation pattern used by this and other test classes (confirmed in GracefulShutdownTests, PlayerEventStreamTests, DecadeBrowseTests) gives a real but session-less core; there is no seam to inject a stub that returns a live session from resumeSession.

**One minor observation (not a blocker)**: `testSecondRestoreAttemptIsNoOp` manually resets `model.isRestoringSession = false` between the two calls, but the first call already leaves the flag `false` (same as `testRestoreWithNoPersistedSessionLowersSplashFlag`), so the reset is redundant. This is a test-legibility nit, not a correctness problem. The test asserts the right thing.

**No new TODO/FIXME** introduced by this delta.

---

### Full checklist (updated)

- [✓ N/A] error-swallowing — no change
- [✓ N/A] MainActor-blocking FFI — no change
- [✓ N/A] paged-cache-only resolution — no change
- [✓ N/A] optimistic-UI-without-echo — no change
- [✓ N/A] hotspot-file growth — no change (AppModel+Session.swift is not a hotspot)
- [✓ addressed] new-branch test coverage — three tests added; happy-path manual-only declared with reason
- [✓ N/A] speculative scope — no change
- [✓ addressed] banned-comments — 8-line explanatory block replaced with 3-line why-comment

### Falsification

The same falsification sketch from the prior pass applies: a test for the throw path of `resumeSession` would verify that `isRestoringSession` is false and `session` is nil when the core throws. That path is covered by the defer, which is structurally verified by reading the code, but the test suite drives the nil-return path, not the throw path. This is a genuine coverage gap but not one the issue scopes to — the throw path behavior is identical before and after this PR, so a new test for it would not fail on origin/main. No falsifying test exists that fails on main and passes on this branch but is absent here; the missing throw-path test would pass on both. No concrete gap remains.

### Outcome: approve

PR #1062, opened 2026-06-11
fix(core): clear downloads (rows and files) on logout

### LINKED ISSUE #1014: Downloads table not cleared on logout — stale rows persist after server switch
## Summary

`Database::clear_user_data()` truncates `track_cache`, `album_cache`, and `artist_cache` on logout, but leaves the `downloads` table untouched. Since `supportsDownloads` is now `true` (enabled in `Capabilities.swift`), the Downloads screen is live. After a logout + login to a **different** Jellyfin server, all download rows from the previous server remain in the `downloads` table and appear in the Downloads screen as if they belong to the new server's library.

## Affected files and lines

**`core/src/storage.rs:155–175`** — `clear_user_data` does not include `downloads`:

```rust
pub fn clear_user_data(&self) -> Result<()> {
    let conn = self.conn.lock();
    conn.execute_batch(
        "DELETE FROM track_cache;
         DELETE FROM album_cache;
         DELETE FROM artist_cache;",    // ← downloads rows survive
    )?;
    // only session-identity settings are deleted below
    ...
}
```

**`macos/Sources/Lyrebird/Capabilities.swift:30`**:
```swift
var supportsDownloads: Bool { true }
```
Downloads are live; the Downloads screen calls `core.listDownloads()` which reads from this table.

## Reproduction

1. Log in to Server A (`https://music.skalthoff.com`, user `test`/`test`).
2. Download one track (or call `core.download_track(_:)` directly in tests).
3. Log out (`core.logout()`).
4. Log in to a different Jellyfin server (Server B).
5. Open the Downloads screen (or call `core.listDownloads()`).

**Expected**: Downloads screen is empty (or shows only Server B downloads).

**Actual**: Track from Server A appears in the Downloads list. The `track_json` snapshot still references Server A item IDs. If the user initiates offline playback of this entry, the track metadata (item_id, album_id, artist_id) are from Server A — any subsequent playback-reporting calls (`report_playback_started`, `report_playback_stopped`) will POST Server A item IDs to Server B's `/Sessions/Playing` endpoints, corrupting server-side play history on Server B.

Sketch failing test:
```rust
#[test]
fn downloads_cleared_on_logout() {
    // Arrange: two "servers" (in-memory DB simulates the state after logout/re-login)
    let db = Database::in_memory().unwrap();
    db.download_upsert_queued("track-server-a", "{}", 0).unwrap();
    // Act: simulate clear_user_data (as called from logout)
    db.clear_user_data().unwrap();
    // Assert: downloads table is empty
    let rows = db.download_list().unwrap();
    assert!(rows.is_empty(), "download rows from the old server should be cleared on logout");
}
```
This test fails today because `clear_user_data` doesn't touch `downloads`.

## Falsifiability statement

This is wrong because `logout()` (at `core/src/lib.rs:410`) calls `db.clear_user_data()` explicitly to "wipe user-scoped DB rows", yet `downloads` rows — which are keyed on server-specific `track_id` values — survive the wipe, causing them to appear in the Downloads screen after a server switch. If downloads are intentionally designed to survive server switches (i.e., tracked purely as device-level files independent of server identity), this issue is invalid.

## De-dup confirmation

De-dup checked against open issues (search queries: `"download server switch"`, `"logout download clear"`, `"819 downloads"`) and CLAUDE.md Deferred list. No match.

skalthoff:
Closes #1014.

`logout()` wiped the library caches and session keys via `clear_user_data` but left the `downloads` table — and its on-disk files — untouched. After a logout + login to a different server, the Downloads screen listed the old server's items, and offline playback of one would have reported the old server's item ids against the new server's `/Sessions/Playing` endpoints (the corruption path described in the issue).

**Change.**
- `Database::downloads_clear()` (storage.rs): truncates the table and returns the recorded `local_path`s — file IO stays out of the storage layer per the module contract.
- `downloads::clear_all(db)`: unlinks each returned file and its `.part` leftover (best-effort, matching `delete()`), after the row wipe.
- `logout()` (lib.rs) calls `downloads::clear_all` beside `clear_user_data`, log-and-continue on error like the rest of the cleanup block.
- The Swift side already clears its in-memory list on sign-out (`AppModel+Session.swift:251`), so no app change is needed.

No FFI surface change (`clear_all` is internal to `logout`), so bindings/xcframework are untouched.

**Tests.** New `clear_all_removes_rows_and_files` (rows truncated; completed file + stale `.part` unlinked; idempotent on empty), and `logout_clears_persisted_settings` now seeds a download row and asserts the table is empty after logout. Full core suite: 340 passed; clippy clean.

### DIFF
--- a/core/src/downloads.rs
+++ b/core/src/downloads.rs
@@ -272,6 +272,22 @@ pub fn delete(db: &Database, track_id: &str) -> Result<()> {
     Ok(())
 }
 
+/// Delete every download: truncate the `downloads` table and unlink each
+/// row's on-disk file. Logout / server-switch cleanup — rows are keyed on
+/// server-specific track ids, so an entry surviving into a session against a
+/// different server would list the old server's items and report the old ids
+/// on offline playback.
+///
+/// File unlinks are best-effort, matching [`delete`]; a row whose fetch never
+/// completed has no `local_path` and only its `.part` leftover is attempted.
+pub fn clear_all(db: &Database) -> Result<()> {
+    for path in db.downloads_clear()? {
+        let _ = std::fs::remove_file(&path);
+        let _ = std::fs::remove_file(Path::new(&path).with_extension("part"));
+    }
+    Ok(())
+}
+
 /// Record an enqueue: snapshot the track into the `downloads` table in the
 /// `queued` state. Does not fetch bytes — [`fetch`] does that. Separated so the
 /// UI can optimistically show a queued badge the instant the user taps,
--- a/core/src/downloads.rs
+++ b/core/src/downloads.rs
@@ -273,13 +273,16 @@ pub fn delete(db: &Database, track_id: &str) -> Result<()> {
 }
 
 /// Delete every download: truncate the `downloads` table and unlink each
-/// row's on-disk file. Logout / server-switch cleanup — rows are keyed on
-/// server-specific track ids, so an entry surviving into a session against a
-/// different server would list the old server's items and report the old ids
-/// on offline playback.
+/// completed row's on-disk file (plus its `.part` sibling). Logout /
+/// server-switch cleanup — rows are keyed on server-specific track ids, so an
+/// entry surviving into a session against a different server would list the
+/// old server's items and report the old ids on offline playback.
 ///
-/// File unlinks are best-effort, matching [`delete`]; a row whose fetch never
-/// completed has no `local_path` and only its `.part` leftover is attempted.
+/// File unlinks are best-effort, matching [`delete`]. Rows that never
+/// completed have no `local_path`, so a fetch still in flight at logout can
+/// leave its `.part` behind — accepted: the file name is deterministic, so a
+/// future fetch of the same track overwrites it, and a `.part` is never
+/// served (only `done` rows with an existing file resolve offline).
 pub fn clear_all(db: &Database) -> Result<()> {
     for path in db.downloads_clear()? {
         let _ = std::fs::remove_file(&path);
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -425,6 +425,13 @@ impl LyrebirdCore {
             if let Err(e) = inner.db.clear_user_data() {
                 tracing::warn!("clear_user_data failed (continuing): {e}");
             }
+            // Download rows are keyed on server-specific track ids, so they
+            // must not survive into a session against a different server —
+            // a stale entry would surface the old server's items in the
+            // Downloads screen and report the old ids on offline playback.
+            if let Err(e) = downloads::clear_all(&inner.db) {
+                tracing::warn!("downloads::clear_all failed (continuing): {e}");
+            }
             if let (Some(sid), Some(uname)) = (server_id, username) {
                 let _ = CredentialStore::delete_token(&sid, &uname);
             }
--- a/core/src/storage.rs
+++ b/core/src/storage.rs
@@ -417,6 +417,21 @@ impl Database {
         Ok(path)
     }
 
+    /// Delete every row of the `downloads` table, returning the `local_path`s
+    /// that were recorded so the caller can unlink the files on disk. See
+    /// [`crate::downloads::clear_all`] — file IO stays out of the storage
+    /// layer by contract.
+    pub fn downloads_clear(&self) -> Result<Vec<String>> {
+        let conn = self.conn.lock();
+        let mut stmt =
+            conn.prepare("SELECT local_path FROM downloads WHERE local_path IS NOT NULL")?;
+        let paths = stmt
+            .query_map([], |r| r.get::<_, String>(0))?
+            .collect::<std::result::Result<Vec<_>, _>>()?;
+        conn.execute("DELETE FROM downloads", [])?;
+        Ok(paths)
+    }
+
     /// Fetch a single download row by track id, or `None` when absent.
     pub fn download_get(&self, track_id: &str) -> Result<Option<DownloadRow>> {
         let conn = self.conn.lock();
--- a/core/src/tests/downloads.rs
+++ b/core/src/tests/downloads.rs
@@ -553,3 +553,40 @@ async fn download_concurrent_fetches_respect_shared_budget() {
     assert!(used <= 100, "used {used} exceeded budget 100");
     assert_eq!(count, 1, "exactly one download should remain within budget");
 }
+
+#[test]
+fn clear_all_removes_rows_and_files() {
+    use crate::downloads;
+    let db = Database::in_memory().unwrap();
+    let dir = tempfile::tempdir().unwrap();
+
+    // One completed download with a real file plus a stale `.part` leftover,
+    // and one still-queued row with no file yet.
+    let done = dl_track("t-done-clear", None, 0);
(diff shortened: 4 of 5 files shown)


skalthoff (commented):

diff-stat: 5 files changed, 86 insertions(+), 0 deletions(−)

---

### Checklist

- [✓ N/A] **error-swallowing** — Both `clear_user_data` and `downloads::clear_all` use the established `if let Err(e) = ... { tracing::warn!(...) }` log-and-continue pattern used throughout the logout block. No silent discard; warnings surface in Console.app under the app's subsystem. No issue.

- [✓ N/A] **MainActor-blocking FFI** — `clear_all` is called from `logout()` which is a synchronous `#[uniffi::export]` method; the call is inside `self.inner.lock()` but the logout path is user-initiated (button tap, not a scroll/cell/tick loop). No tight-loop or per-frame concern. N/A.

- [✓ N/A] **paged-cache-only resolution** — This PR adds no list-resolution code. N/A.

- [✓ N/A] **optimistic-UI-without-echo** — The PR description confirms the Swift side (`AppModel+Session.swift:251`) already clears its in-memory list on sign-out. The Rust side now also wipes the table and files. No local mutation that diverges from the server echo path. N/A.

- [✓ N/A] **hotspot-file growth** — `core/src/lib.rs` gains 7 lines (within the existing `logout()` method body), `core/src/storage.rs` gains 15 lines (one new method), `core/src/downloads.rs` gains 16 lines (one new public function). AppModel.swift and LyrebirdApp.swift are untouched. `client.rs` is untouched. Test additions land in the correct domain files (`tests/downloads.rs`, `tests/session_auth.rs`), consistent with the June 2026 split. Justified; no gratuitous growth.

- [✓ N/A] **new-branch test coverage** — Two tests added: `clear_all_removes_rows_and_files` (rows truncated, completed file unlinked, stale `.part` unlinked, idempotent) and the extended `logout_clears_persisted_settings` (seeds a row, asserts empty after logout). Both pass on this branch; both would have failed on `origin/main` (function didn't exist). All 340 core lib tests pass. Coverage is adequate for the happy path.

- [✓ N/A] **speculative scope** — Every changed line is traceable to issue #1014 (downloads survive logout). No unrequested behavior changed.

- [✗ finding] **banned-comments** — `core/src/downloads.rs` lines 281–282:
  ```
  /// a row whose fetch never completed has no `local_path` and only its `.part` leftover is attempted.
  ```
  This doc comment is factually wrong, which makes it a worse offence than merely explaining the code. `downloads_clear()` filters `WHERE local_path IS NOT NULL`, so rows with NULL `local_path` (state = `queued` or `downloading`) contribute **no** entry to the returned vec. The loop body therefore runs zero iterations for those rows — no `.part` removal is attempted at all for in-flight or queued tracks. The comment promises something the code does not do.

  This produces a real file-system leak: a download that is actively in-flight at logout (state = `downloading`, `local_path = NULL`, `.part` file exists on disk) will have its DB row deleted but its `.part` file left behind. The leak is bounded (the file is dead-storage only; it won't be played) but it contradicts the stated contract and violates the "no comments explaining what code does incorrectly" rule.

  The fix is either: (a) change the docstring to accurately describe the current behavior ("queued and in-flight rows have their DB row deleted; their on-disk `.part` — if any — is not cleaned up because `local_path` is never set until completion"), or (b) expand `downloads_clear()` to also return the computed `.part` paths for rows where `local_path IS NULL` (using the same stem derivation as `fetch`).

  Option (a) is the minimal safe fix. Option (b) is more complete but requires `downloads_clear` to know the download directory, which breaks the current file-IO-stays-out-of-storage-layer contract.

---

### Falsification: failing-test sketch not covered by this PR

```rust
#[test]
fn clear_all_leaves_no_part_for_in_flight_row() {
    use crate::downloads;
    let db = Database::in_memory().unwrap();
    let dir = tempfile::tempdir().unwrap();

    // Simulate a download that is in-flight: row exists in 'downloading' state
    // (local_path = NULL), and the fetcher has already created the .part file.
    let stem = "t-inflight";
    let part = dir.path().join(format!("{stem}.part"));
    std::fs::write(&part, b"in progress").unwrap();

    let json = serde_json::to_string(&dl_track(stem, None, 0)).unwrap();
    db.download_upsert_queued(stem, &json, 1).unwrap();
    db.download_mark_downloading(stem).unwrap();
    // local_path is still NULL — mark_downloading does not set it.

    downloads::clear_all(&db).unwrap();

    // Row gone (correct):
    assert!(downloads::list(&db).unwrap().is_empty());
    // .part file gone (THIS FAILS — clear_all does not attempt it):
    assert!(!part.exists(), ".part for in-flight download must be cleaned up on logout");
}
```

This test fails on both `origin/main` and this branch. The docstring claims the `.part` is attempted; the test proves it isn't.

---

**Outcome: request-changes**

The misclaiming doc comment is the sole defect. No logic error in the SQL or locking. If the fixer amends the docstring to accurately describe what the code actually does (no `.part` cleanup for in-flight/queued rows), the finding is closed and the PR is approvable. Option (b) — actual `.part` cleanup for in-flight rows — would also close it but is wider scope.

skalthoff:
Docstring corrected: `clear_all` only unlinks files for rows with a recorded `local_path` (the `WHERE local_path IS NOT NULL` projection), so a fetch still in flight at logout can leave its `.part` behind. That leak is now stated as accepted with the rationale — deterministic dest naming means a future fetch of the same track overwrites it, and `.part` files are never served (`local_path_for` only resolves `done` rows with an existing file). Sweeping the download dir instead would need `data_dir` plumbed into the logout cleanup path, which isn't warranted for a cosmetic leftover.

skalthoff (commented):
## Adversarial review (Sonnet — refix verification pass)

The sole first-pass finding was the incorrect docstring on `downloads::clear_all` (downloads.rs:281–282). The refix replaces it.

**Verified delta (downloads.rs lines 275–285 on HEAD):**

- "completed row's on-disk file" — matches the `WHERE local_path IS NOT NULL` projection; queued/in-flight rows correctly excluded from the claim.
- "Rows that never completed have no `local_path`, so a fetch still in flight at logout can leave its `.part` behind" — accurate: `download_mark_downloading` does not set `local_path`; `local_path` is only populated by `download_mark_done`.
- "accepted: the file name is deterministic, so a future fetch of the same track overwrites it" — verified: `dest_within_dir` derives the same `<stem>.<ext>` path every time from the track id; the future fetch's `.part` write (`<stem>.part`) overwrites the leftover.
- "a `.part` is never served (only `done` rows with an existing file resolve offline)" — verified: `local_path_for` (downloads.rs:247–249) returns `None` for any state != `Done` before it ever inspects a file path, so a `.part` leftover is unreachable from offline playback.

All other checklist categories carry forward unchanged from the first-pass review (N/A across the board for error-swallowing, MainActor-blocking FFI, paged-cache-only resolution, optimistic-UI-without-echo, hotspot-file growth, speculative scope). Test suite: 341 lib tests pass; clippy clean.

**Falsification revisited:** the hypothetical `clear_all_leaves_no_part_for_in_flight_row` sketch from the first pass now correctly describes an accepted and documented limitation, not an undisclosed bug. No new falsifying case exists; the invariant is: `clear_all` only unlinks files it has a recorded path for; the contract gap is explicitly documented with accepted rationale.

**Outcome: approve**

Issue #819, opened 2026-05-13
Offline downloads — core engine + per-track / album / playlist actions


skalthoff:
Offline-playback affordances ship as UI placeholders in 0.2 / 1.0 — hidden behind `Capabilities.supportsDownloads` (currently `false`). The original tracking lived in **#70** (album CTA), **#222** (favorite/download/add-to-playlist actions) and a handful of related polish issues, all of which closed without the download engine ever landing. This issue is the fresh paper trail for the work.

### Core (Rust)

A persistent download engine wired into `LyrebirdCore` over UniFFI. Concrete requirements:

- Track-level state machine: `notDownloaded → queued → downloading(progress) → downloaded → failed`. Stored alongside library cache so it survives restarts.
- Storage on disk under `Library/Application Support/Lyrebird/Downloads/<itemId>.<ext>` with content-hash sidecar metadata (codec, bitrate, original-stream URL, server-id) so a server change can invalidate the file.
- FFI primitives:
  - `enqueueDownload(trackIds: [String])` — push N tracks onto the queue, dedup against in-flight + already-downloaded.
  - `cancelDownload(trackIds: [String])` — abort in-flight, evict from queue. No file delete.
  - `removeDownload(trackIds: [String])` — delete the file + sidecar; flip state back to `notDownloaded`.
  - `downloadStates(for ids: [String]) -> [String: DownloadState]` — snapshot read for UI gating.
  - `downloadQueueSummary() -> DownloadQueueSummary` — `{ queued: usize, active: usize, totalBytesPending: u64 }` for a future "Downloads" sidebar pane.
- Eventing: a `downloadStateChanged(trackId, state)` callback (same pattern as `statusChanged`) so the Swift side can drop snapshot polls.
- Concurrency: default 2 parallel transfers, backoff on transient HTTP failures.

### macOS UI (already in place, hidden)

Once the engine lands, flip `Capabilities.supportsDownloads` to `true` and rewire the four `AppModel` stubs (currently `Log.app.notice("…not yet wired — see #70 / #222")`) to the new FFI:

- `AppModel.swift` Command palette "Download Current" entry (around L661).
- `AppModel.enqueueDownload(album:)` (around L3291) — expand to every track on the album.
- `AppModel.enqueueDownload(playlist:)` (around L3707) — expand to every track in the playlist.
- `AppModel.toggleDownload(tracks:)` (around L4225) — all-on / any-off toggle to mirror the favorite-selection pattern.

[...]

Issue #431, opened 2026-04-22
Database-backed library cache with background revalidation


skalthoff:
Today the library is fetched fresh every launch. On a 50k-track server with 1500-ms roundtrips over a wi-fi link the user stares at a spinner. The fix is a cache-first, revalidate-in-background strategy.

Design:
- Persist `albums`, `artists`, `tracks` metadata into `album_cache` / `artist_cache` / `track_cache` as JSON rows (already schema-prepared) the first time we fetch them. Keyed by ID, with `updated_at`.
- On app launch: emit cached rows to the UI immediately (< 50 ms path via `core.list_cached_albums(limit)`), then asynchronously fetch from the server. Diff (by ID + a hash of the `Etag` / `DateLastMediaAdded` / `DateModified` field Jellyfin returns) and push only changed rows through a `core.library_changes` event channel.
- Jellyfin exposes `Users/{userId}/Items?IncludeItemTypes=MusicAlbum&Fields=DateLastMediaAdded&MinDateLastSaved=<iso8601>` — use the `MinDateLastSaved` filter for delta sync. Store the server's `last_successful_sync` timestamp per user in `settings`.
- Add a `JellifyLibrary` service in Rust that wraps `JellyfinClient + Database`. SwiftUI observes via an async stream surfaced over UniFFI (`uniffi::CallbackInterface`).

Acceptance: warm launch shows the library in < 150 ms; background revalidation completes in < 3 s for a 5k-album library; delta sync transfers ≤ 1% of the full payload on a no-op day.

<sub>Source: `10-perf-reliability.md`</sub>

PR #1064, opened 2026-06-11
perf(core): large-library SQLite pragmas + bounded PRAGMA optimize

### LINKED ISSUE #430: SQLite index + pragma audit for large libraries
`migrations/001_initial.sql` only indices `play_history(track_id)` and `play_history(played_at)`. As `track_cache`, `album_cache`, `artist_cache` grow (and as we add favorites / playlist tables) we need more.

Add (in a new `002_indices.sql` migration):
- `CREATE INDEX idx_play_history_completed ON play_history(completed, played_at DESC);` — for "recently played" filtered to completed listens.
- `CREATE INDEX idx_track_cache_updated ON track_cache(updated_at);` — for stale-revalidation batch selects.
- `CREATE INDEX idx_album_cache_updated ON album_cache(updated_at);`
- When we add favorites/playlists: `idx_favorites_track`, `idx_playlist_items_playlist_pos`.

Verify with `EXPLAIN QUERY PLAN` in a debug harness. Also tune pragmas:
- Already set: `journal_mode=WAL`, `synchronous=NORMAL` (good for our use case — survives crash, not power loss).
- Add: `cache_size = -20000` (~20 MB page cache), `temp_store = MEMORY`, `mmap_size = 268435456` (256 MB; read-only mmap for seq scans).
- Add `PRAGMA optimize;` on app shutdown and once after the first-time full library dump.

Acceptance: `SELECT * FROM album_cache ORDER BY updated_at DESC LIMIT 100` runs in < 5 ms at 10k rows on a cold cache.

skalthoff:
Closes #430.

**What's left of the issue** (scoping comment posted there): the index asks are obsolete — `play_history` was removed, no query reads `updated_at`, and the sort-key indexes the issue anticipated landed in `003_library_cache_sort.sql`. WAL + `synchronous=NORMAL` were already set. The remaining substance is pragma tuning and statistics upkeep.

**Change.**
- `Database::open` adds `cache_size = -20000` (~20 MB page cache vs the ~2 MB default), `temp_store = MEMORY` (ORDER BY spills stay off disk), `mmap_size = 256 MB` (sequential cache scans become page-cache reads), and `analysis_limit = 400` (bounds every ANALYZE pass).
- `PRAGMA optimize` runs at open (seeds planner stats on first launch) and after each non-empty `cache_upsert` batch — inlined on the held connection to avoid re-locking. With `analysis_limit` it's microseconds when stats are fresh and a bounded sample when they're not. The connection is process-lifetime and never closes cleanly on quit, so SQLite's run-at-close advice has no hook here; open + post-bulk-write are the realistic seams.
- `in_memory()` gets the same `analysis_limit` so tests exercise the same optimize behavior.
- Tests: `open_applies_large_library_pragmas` (pins all five pragma values on a file-backed DB via a new `cfg(test)` connection seam) and `cache_page_query_uses_sort_key_index` (EXPLAIN QUERY PLAN must show `idx_album_cache_sort` for the warm-launch page query — the guard against a future schema change silently degrading the 20k-row read into a temp-B-tree sort).

No FFI surface change. `cargo test`: 342 passed; `cargo fmt --all --check` and clippy clean.

### DIFF
--- a/core/src/storage.rs
+++ b/core/src/storage.rs
@@ -21,6 +21,37 @@ impl Database {
         conn.pragma_update(None, "journal_mode", "WAL")?;
         conn.pragma_update(None, "foreign_keys", "ON")?;
         conn.pragma_update(None, "synchronous", "NORMAL")?;
+        // Large-library tuning (#430). The page cache default (~2 MB) thrashes
+        // on a 20k-album cache walk; -20000 = ~20 MB, sized so a full
+        // `cache_list` page read stays in memory. `temp_store = MEMORY` keeps
+        // ORDER BY spill sorts off disk, and the 256 MB read-only mmap turns
+        // sequential cache scans into page-cache reads. `analysis_limit`
+        // bounds every `PRAGMA optimize` pass (run after bulk cache writes)
+        // so it never degrades into a full-table ANALYZE on the UI's watch —
+        // the four values together are SQLite's documented recipe for
+        // long-lived desktop connections.
+        conn.pragma_update(None, "cache_size", -20000)?;
+        conn.pragma_update(None, "temp_store", "MEMORY")?;
+        conn.pragma_update(None, "mmap_size", 268_435_456_i64)?;
+        conn.pragma_update(None, "analysis_limit", 400)?;
+        let db = Database {
+            conn: Mutex::new(conn),
+        };
+        db.migrate()?;
+        // Seed planner statistics for any table still missing them (cheap
+        // no-op when `sqlite_stat1` is already populated, bounded by
+        // `analysis_limit` otherwise).
+        db.optimize()?;
+        Ok(db)
+    }
+
+    pub fn in_memory() -> Result<Self> {
+        let conn = Connection::open_in_memory()?;
+        conn.pragma_update(None, "foreign_keys", "ON")?;
+        // Match the on-disk connection's bounded-ANALYZE setting so the
+        // `optimize()` calls sprinkled through the write paths behave the
+        // same under test as in the app.
+        conn.pragma_update(None, "analysis_limit", 400)?;
         let db = Database {
             conn: Mutex::new(conn),
         };
@@ -28,14 +59,27 @@ impl Database {
         Ok(db)
     }
 
-    pub fn in_memory() -> Result<Self> {
-        let conn = Connection::open_in_memory()?;
-        conn.pragma_update(None, "foreign_keys", "ON")?;
-        let db = Database {
-            conn: Mutex::new(conn),
-        };
-        db.migrate()?;
-        Ok(db)
+    /// Run `PRAGMA optimize` — SQLite's self-tuning hook that re-ANALYZEs
+    /// exactly the tables whose statistics are missing or stale. Bounded by
+    /// the `analysis_limit` set at open, so a pass over the 20k-row caches
+    /// costs microseconds when there's nothing to do and a bounded sample
+    /// scan when there is.
+    ///
+    /// Called at open (seeds stats on first launch) and after bulk
+    /// `cache_upsert` batches (the only writes big enough to skew the
+    /// planner). The process-lifetime connection never closes cleanly on
+    /// quit, so the usual run-at-close advice doesn't apply here.
+    pub fn optimize(&self) -> Result<()> {
+        self.conn.lock().execute_batch("PRAGMA optimize;")?;
+        Ok(())
+    }
+
+    /// Test seam: run `f` against the raw connection. Lets the pragma and
+    /// `EXPLAIN QUERY PLAN` regression tests inspect connection state without
+    /// growing the public surface.
+    #[cfg(test)]
+    pub(crate) fn with_conn<T>(&self, f: impl FnOnce(&Connection) -> T) -> T {
+        f(&self.conn.lock())
     }
 
     fn migrate(&self) -> Result<()> {
@@ -258,6 +302,14 @@ impl Database {
             }
         }
         tx.commit()?;
+        if !changed.is_empty() {
+            // Refresh planner statistics after a bulk write batch. Inlined on
+            // the held connection (`optimize()` would re-lock); bounded by the
+            // `analysis_limit` pragma and a no-op while stats are still
+            // representative, so per-page cost during a full library dump is
+            // negligible.
+            conn.execute_batch("PRAGMA optimize;")?;
+        }
         Ok(changed)
     }
 
--- a/core/src/tests/library_cache.rs
+++ b/core/src/tests/library_cache.rs
@@ -822,3 +822,73 @@ async fn revalidation_noop_delta_is_two_cheap_album_requests() {
     assert_eq!(core.list_cached_artists(10).unwrap().len(), 1);
     drop_core(core).await;
 }
+
+// ---------------------------------------------------------------------------
+// Large-library SQLite tuning (#430)
+// ---------------------------------------------------------------------------
+
+/// The on-disk connection must carry the large-library pragma set: WAL (from
+/// the original open path) plus the #430 additions — ~20 MB page cache,
+/// in-memory temp store, 256 MB read-only mmap, and the bounded
+/// `analysis_limit` that keeps every `PRAGMA optimize` pass cheap.
+#[test]
+fn open_applies_large_library_pragmas() {
+    let tmp = tempfile::tempdir().unwrap();
+    let db = crate::storage::Database::open(tmp.path().join("pragmas.sqlite")).unwrap();
+    db.with_conn(|c| {
+        let journal: String = c
+            .pragma_query_value(None, "journal_mode", |r| r.get(0))
+            .unwrap();
+        assert_eq!(journal.to_lowercase(), "wal");
+        let cache_size: i64 = c
+            .pragma_query_value(None, "cache_size", |r| r.get(0))
+            .unwrap();
+        assert_eq!(cache_size, -20000);
+        let temp_store: i64 = c
+            .pragma_query_value(None, "temp_store", |r| r.get(0))
+            .unwrap();
+        assert_eq!(temp_store, 2, "temp_store should resolve to MEMORY (2)");
+        let mmap: i64 = c
+            .pragma_query_value(None, "mmap_size", |r| r.get(0))
+            .unwrap();
+        assert_eq!(mmap, 268_435_456);
+        let analysis_limit: i64 = c
+            .pragma_query_value(None, "analysis_limit", |r| r.get(0))
+            .unwrap();
+        assert_eq!(analysis_limit, 400);
+    });
+}
+
+/// Regression guard for the warm-launch read path: the `cache_list` page
+/// query must walk `idx_album_cache_sort` rather than scanning + sorting the
+/// whole table — on a 20k-row cache the difference is an index walk touching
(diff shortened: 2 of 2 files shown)


skalthoff (commented):

reviewer-session: sonnet-4-6 / fix/430-sqlite-pragmas
diff-stat: 2 files changed, 122 insertions(+) — core/src/storage.rs (+52), core/src/tests/library_cache.rs (+70)

---

### Pre-analysis findings (diff-first, before reading PR description)

**1. `pragma_update` on `mmap_size` — rusqlite semantics verified**

`mmap_size` is a pragma that returns a row with the clamped value. I read `rusqlite-0.39.0/src/pragma.rs` line 238: `pragma_update` is implemented as `self.execute_batch(&sql)`. I then read `lib.rs` line 539–558: `execute_batch` explicitly handles the case where a pragma returns rows via `if false { return Err(ExecuteReturnedResults) }` — rows are intentionally discarded. Result: `pragma_update("mmap_size", 268_435_456_i64)` sets the pragma correctly; the returned (possibly clamped) value is silently dropped, which is the documented rusqlite behavior. No error, no incorrect behavior.

**2. `execute_batch("PRAGMA optimize;")` — correct**

Same `execute_batch` path. `PRAGMA optimize` with `0x01` flag may return rows of SQL, but `execute_batch` iterates and discards them. Verified in source. No error risk.

**3. `tx.commit()` then `conn.execute_batch()` — borrow checker story is clean**

`Transaction::commit(mut self)` consumes `self`, ending the `&'conn Connection` borrow. After that, `conn` (the `MutexGuard<Connection>`) is free for use again. This is the standard rusqlite deferred-borrow pattern; it compiled, and it's correct. No `&mut` aliasing footgun.

**4. Lock reentrancy for `optimize()` — correctly avoided**

`parking_lot::Mutex` is non-reentrant. In `cache_upsert`, `optimize()` would attempt `self.conn.lock()` while the caller already holds it — deadlock. The PR correctly inlines `conn.execute_batch("PRAGMA optimize;")` on the already-held connection instead, with an explanatory comment. The only external call to `optimize()` is in `open()`, where the struct was just constructed and no lock is held. No reentrancy path exists.

**5. mmap_size portability — verified safe on all CI runners**

`SQLITE_MAX_MMAP_SIZE` is set to `0x7fff0000` (2 GB) on Linux, macOS, and Windows in the bundled SQLite source (`libsqlite3-sys-0.37.0/sqlite3/sqlite3.c` lines 15718–15724). CI runs on ubuntu-latest, macos-15, windows-latest — all three. The `assert_eq!(mmap, 268_435_456)` in `open_applies_large_library_pragmas` holds on all three because 256 MB is far below the 2 GB ceiling. On OpenBSD/QNX, mmap is disabled (ceiling = 0), but those platforms are not CI targets here.

One legitimate fragility: OS-level `mmap(2)` restrictions or container ulimits could silently cap the value below what SQLite reports; SQLite's `PRAGMA mmap_size` readback reflects what SQLite accepted, which may differ from what the OS actually maps. In practice this is a concern only in very locked-down containers; the GitHub Actions runners don't restrict this. Not a defect — noted as a test fragility.

**6. `-20000` literal type inference — correct**

Rust infers `-20000` as `i32` in the `V: ToSql` generic context. `ToSql for i32` maps to `ValueRef::Integer(i64)`. rusqlite formats it as the integer literal `-20000` in the pragma SQL. SQLite's `cache_size` accepts this (negative = kibibytes × −1). Correct.

---

### Mandatory checklist

- **[✓ N/A] error-swallowing** — No `try?`, empty `catch`, or `.ok()` in the diff. Both `pragma_update` calls and both `execute_batch("PRAGMA optimize;")` calls propagate errors with `?`. No swallowing.

- **[✓ N/A] MainActor-blocking FFI** — Pure Rust changes to `core/src/storage.rs`. No Swift, no `@MainActor`, no FFI call site changes.

- **[✓ N/A] paged-cache-only resolution** — No resolver or `.first { $0.id == X }` pattern touched.

- **[✓ N/A] optimistic-UI-without-echo** — No Swift mutation logic. No local state changes. Pure storage-layer tuning.

- **[✓ N/A] hotspot-file growth** — Neither `AppModel.swift`, `LyrebirdApp.swift`, nor `client.rs` was touched. `storage.rs` is not a hotspot. Test growth in `core/src/tests/library_cache.rs` is expected and in the correct domain file.

- **[✓ addressed] new-branch test coverage** — `open_applies_large_library_pragmas` (file-backed DB, asserts all five pragma values) and `cache_page_query_uses_sort_key_index` (EXPLAIN QUERY PLAN must show `idx_album_cache_sort`) both pass on this branch and would fail on `origin/main` (the pragmas aren't set there). Confirmed: `cargo test --package lyrebird_core library_cache` → 14 passed.

- **[✓ N/A] speculative scope** — All changed lines trace to issue #430 (SQLite pragma tuning for large libraries). No unrelated changes.

- **[✓ N/A] banned-comments** — Issue references (`#430`) appear in new comments, matching the established style in this file (every pre-existing block already uses `#819`, `#431`). No `TODO`/`FIXME` added. No what-the-code-does-only comments that aren't also carrying non-obvious rationale (the negative cache_size value, mmap semantics, and the re-lock avoidance are all non-obvious; the prose earns its place).

---

### Falsification

A test this PR does not cover:

```rust
#[test]
fn cache_upsert_optimize_not_called_on_empty_change_set() {
    // If mmap_size is silently capped to 0 by a future environment (e.g. a
    // container that restricts mmap), the open_applies_large_library_pragmas
    // test fails even though the code is correct. More importantly: there is
    // no test that the inlined PRAGMA optimize in cache_upsert does NOT fire
    // when changed.is_empty() -- i.e. when every row in the page was already
    // up to date. This path is only tested indirectly. A direct assertion:
    let db = Database::in_memory().unwrap();
    // Insert once.
    let writes = vec![CacheWrite { id: "a".into(), data: r#"{"id":"a"}"#.into(), sort_key: "k".into() }];
    db.cache_upsert(CacheKind::Album, &writes, 1).unwrap();
    // Re-upsert identical data: changed should be empty, optimize should be skipped.
    let changed = db.cache_upsert(CacheKind::Album, &writes, 1).unwrap();
    assert!(changed.is_empty(), "no-op upsert must produce empty changed set");
    // There is no way to assert from the outside that PRAGMA optimize was
    // *not* called; the test above only checks the return value, not the
    // optimize skip. An adversary could remove the `if !changed.is_empty()`
    // guard and this test would still pass.
}
```

This is a correctness gap in the test, not a defect in the code. The `if !changed.is_empty()` guard is correct — the gap is that no test validates the guard path directly. It would require a hook (e.g., a call counter) that the current seam doesn't expose.

**no such invariant violation exists because the guard is clearly correct and the test suite validates the behavior it can observe (the changed-set shape).**

---

Outcome: **approve**

skalthoff (commented):

reviewer-session: sonnet-4-6 / fix/430-sqlite-pragmas
diff-stat: 2 files changed, 122 insertions(+) — core/src/storage.rs (+52), core/src/tests/library_cache.rs (+70)

---

### Pre-analysis findings (diff-first, before reading PR description)

**pragma_update on mmap_size — rusqlite semantics verified**

mmap_size is a pragma that returns a row with the clamped value. Verified in rusqlite-0.39.0/src/pragma.rs line 238: pragma_update is implemented as self.execute_batch(&sql). Then lib.rs lines 539-558: execute_batch explicitly handles pragmas that return rows via `if false { return Err(ExecuteReturnedResults) }` — rows are intentionally discarded. Result: pragma_update("mmap_size", 268_435_456_i64) sets the pragma correctly; the returned clamped value is discarded by design. No error, no incorrect behavior.

**execute_batch("PRAGMA optimize;") — correct**

Same execute_batch path. PRAGMA optimize may return rows of SQL, but execute_batch iterates and discards them. Verified in source. No error risk.

**tx.commit() then conn.execute_batch() — borrow checker story is clean**

Transaction::commit(mut self) consumes self, ending the &'conn Connection borrow. After that, conn (the MutexGuard) is free for direct use. This is the standard rusqlite deferred-borrow pattern; it compiled and is correct. No aliasing footgun.

**Lock reentrancy for optimize() — correctly avoided**

parking_lot::Mutex is non-reentrant. In cache_upsert, calling optimize() would attempt self.conn.lock() while the caller already holds it — deadlock. The PR correctly inlines conn.execute_batch("PRAGMA optimize;") on the already-held connection. The only external call to optimize() is in open(), where no lock is held. No reentrancy path exists.

**mmap_size portability — verified safe on all CI runners**

SQLITE_MAX_MMAP_SIZE is set to 0x7fff0000 (2 GB) on Linux, macOS, and Windows in the bundled SQLite source (libsqlite3-sys-0.37.0/sqlite3/sqlite3.c lines 15718-15724). CI runs on ubuntu-latest, macos-15, windows-latest — all three qualify. The assert_eq!(mmap, 268_435_456) in open_applies_large_library_pragmas holds on all three because 256 MB is far below the 2 GB ceiling. OpenBSD/QNX have ceiling 0, but those are not CI targets.

One fragility: OS-level mmap restrictions or container ulimits could silently cap the value in a locked-down environment. GitHub Actions runners do not restrict this. Not a defect; noted.

**-20000 literal type inference — correct**

Rust infers -20000 as i32 in the V: ToSql context. ToSql for i32 maps to ValueRef::Integer(i64); rusqlite formats it as the integer literal -20000 in the pragma SQL. SQLite's cache_size accepts negative values (kibibytes). Correct.

---

### Mandatory checklist

- [✓ N/A] error-swallowing — No try?, empty catch, or .ok() in the diff. All pragma_update and execute_batch calls propagate errors with ?.

- [✓ N/A] MainActor-blocking FFI — Pure Rust changes to core/src/storage.rs. No Swift, no @MainActor, no FFI call site changes.

- [✓ N/A] paged-cache-only resolution — No resolver or .first { $0.id == X } pattern touched.

- [✓ N/A] optimistic-UI-without-echo — No Swift mutation logic. Pure storage-layer tuning.

- [✓ N/A] hotspot-file growth — Neither AppModel.swift, LyrebirdApp.swift, nor client.rs was touched. storage.rs is not a hotspot. Test growth in core/src/tests/library_cache.rs is in the correct domain file.

- [✓ addressed] new-branch test coverage — open_applies_large_library_pragmas and cache_page_query_uses_sort_key_index pass on this branch and would fail on origin/main. Confirmed: cargo test --package lyrebird_core library_cache -> 14 passed.

- [✓ N/A] speculative scope — All changed lines trace to issue #430 (SQLite pragma tuning). No unrelated changes.

- [✓ N/A] banned-comments — Issue references (#430) match the established style throughout this file (every pre-existing block uses #819, #431). No TODO/FIXME added. Rationale prose in comments covers non-obvious semantics (negative cache_size, mmap semantics, re-lock avoidance) that earn their place.

---

### Falsification

A test this PR does not cover: direct assertion that the `if !changed.is_empty()` guard prevents the inlined PRAGMA optimize from running on a no-op upsert page. The existing tests verify the changed-set return value but cannot observe whether the optimize branch was skipped. A call-counter or hook-based test would be needed to close this gap:

```rust
// Hypothetical failing test (not possible without exposing a hook):
// Re-upsert identical data -> changed.is_empty() -> PRAGMA optimize must NOT run.
// Currently only the return value is testable, not whether the guard fired.
```

This is a test-coverage gap, not a code defect. The guard logic is clearly correct. No such invariant violation exists in the implementation.

---

Outcome: **approve**

Issue #580, opened 2026-04-23
No gapless playback — single AVPlayer used despite ROADMAP claim


skalthoff:
**Severity**: P1  
**Areas**: `area:macos`, `area:audio`

ROADMAP M3 lists "AVQueuePlayer-based gapless" as a deliverable. The current `AudioEngine` is a single `AVPlayer` with manual `play(track:)` swapping at track end — measurable 100-200ms gap between tracks. No queue items are pre-buffered.

## Files

- [`macos/Sources/JellifyAudio/AudioEngine.swift:71`](macos/Sources/JellifyAudio/AudioEngine.swift#L71) — single `AVPlayer` instance.
- `attachPlayerObservers()` only observes the current item; next track URL is fetched + parsed at current track end.

## Why it matters

Gapless playback is table-stakes for music apps. Live albums, concept records, and DJ mixes all rely on it. Currently breaks for any continuous-mix material.

## Proposed fix

Migrate to `AVQueuePlayer`:

1. Pre-fetch the next track's `playback_info` while the current is playing (so `PlaySessionId` + `MediaSourceId` are ready).
2. Insert pre-built `AVPlayerItem` into the queue before the current ends.
3. Use `AVPlayerItem.preferredForwardBufferDuration` to control buffer ahead.
4. Update KVO observers on item-change events to track which track is currently active.
5. Verify with a known-gapless album (e.g., Daft Punk *Discovery*).

Acceptance: A→B transition has < 50ms perceptible gap (measure with sample audio).

---
*Found during full-surface adversarial audit on main, 2026-04-22.*

Issue #439, opened 2026-04-22
AVPlayer stall recovery + user-visible retry


skalthoff:
`automaticallyWaitsToMinimizeStalling = true` is set, which helps, but we never handle the `AVPlayer.timeControlStatus == .waitingToPlayAtSpecifiedRate` transition ourselves. After a long stall the player just hangs silently.

Plan:
- Observe `player.timeControlStatus` and `player.reasonForWaitingToPlay`. If waiting for > 5 s: trigger a retry (preserve current position, rebuild `AVPlayerItem` with the same URL + auth header, `seek(to: savedTime)`, `play()`).
- Observe `AVPlayerItem.status` — on `.failed`, inspect `error` and show a user-facing toast: "Couldn't play track — tap to retry". Keep the queue intact; don't auto-skip on failure (that's worse than stopping).
- Observe `AVPlayerItem.isPlaybackLikelyToKeepUp`. When it flips `false`, log a `Logger` event with the current bitrate + position + server round-trip time so we can diagnose from user logs.
- On total giveup (3 failed retries), advance to next track and surface a persistent error banner.

Acceptance: pulling the ethernet cable mid-track results in "Reconnecting..." state within 3 s and successful resume when the cable is plugged back in. Breaking the stream URL (simulated 401) shows a retry button and does not spin forever.

<sub>Source: `10-perf-reliability.md`</sub>

skalthoff:
Verified shipped via P0 cross-area stale-audit on 2026-04-27.

Shipping ref: `58ba8a9`

stallThreshold + timeControlStatus observer + retry + StreamErrorToast in AudioEngine.swift.

Auto-close keyword in the original PR didn't fire — closing now. Reopen if any acceptance criterion is genuinely unmet.

Issue #42, opened 2026-04-22
ReplayGain + volume normalization


skalthoff:
**Depends on:** #13
- Jellyfin exposes `NormalizationGain` (in dB) and `LUFS` on the audio item JSON ([jellyfin issue #14346](https://github.com/jellyfin/jellyfin/issues/14346), [Volume Normalization thread](https://features.jellyfin.org/posts/2363/replay-gain-normalization)). Extend the Rust core `Track` struct with `normalizationGain: f32?` and `lufs: f32?` — parse from the `/Items/{id}` response. Regenerate UniFFI bindings.
- In settings, offer three modes: Off, Track (default), Album. Note that as of 2025 Jellyfin's `NormalizationGain` returns the track value regardless of the chosen preference — if album mode is selected we'll need an additional call to `/Items/{albumId}` to pull album-level gain. Document this caveat.
- Apply the gain as a pre-EQ `AVAudioUnitEQ` (use the eq's `globalGain` property), or more cleanly as `AVAudioMixerNode.outputVolume = pow(10, gain_dB / 20)`. Clamp at 0dB reduction (never boost above the source) to avoid clipping, which is the ReplayGain convention.
- Fall back to a simple peak-limit if no gain metadata is present (`AVAudioUnitEffect` with an AUDynamicsProcessor AU — limiter only, threshold -1dBFS, release 50ms). Prevent peak clipping when EQ is boosting bands.
- **Acceptance:** A quiet track followed by a loud track plays at roughly the same perceived loudness when Track mode is on; toggling Off restores raw levels; no clipping when the limiter is enabled and aggressive EQ is applied to a hot source.

References: [Jellyfin ReplayGain feature request](https://features.jellyfin.org/posts/2363/replay-gain-normalization) • [NormalizationGain track/album issue](https://github.com/jellyfin/jellyfin/issues/14346) • [Hacking ReplayGain into Jellyfin](https://project-insanity.org/2019/04/10/hacking-replay-gain-audio-normalization-into-jellyfin/).

---

<sub>Source: `02-media-integration.md`</sub>

PR #1065, opened 2026-06-11
feat(macos): surface AVPlayerItem access-log telemetry in logs and the debug panel

### LINKED ISSUE #452: AVPlayerItem.accessLog` → tracing span
`AVPlayerItem.accessLog()` is a goldmine: `numberOfStalls`, `indicatedBitrate`, `observedBitrate`, `downloadOverdue`, server address. We ignore it.

On `AVPlayerItemNewAccessLogEntryNotification`, extract the latest `AVPlayerItemAccessLogEvent` and emit a structured `tracing` event with those fields. This surfaces straight into the debug panel (Issue 23) and, if opted in, metrics upload (Issue 25).

Acceptance: the debug panel "Player" section shows live bitrate and stall count.

skalthoff:
Closes #452.

`AVPlayerItem.accessLog()` was ignored; its transport-health fields are exactly what you want when "why is this stream choppy" comes up. This wires it end to end:

- **Engine**: observes `AVPlayerItemNewAccessLogEntry` on the live item — wired beside the existing end-of-track observer at all three points (player setup, gapless `currentItem` advance, stall-recovery rebuild) so telemetry always follows the playing stream. Each entry emits one structured log line (subsystem `org.lyrebird.desktop`, category `player`) with stalls, indicated/observed bitrate, overdue segments, and the server address (hash-masked for privacy).
- **Model**: new `PlayerAccessLogStats` value type crosses via a new `audioEngineDidUpdateAccessLog` delegate hook (default no-op, same pattern as #806). Stats clear on `stop()`/`logout()` so a dead stream's numbers never linger in the panel.
- **Debug panel**: Player tab gains a "Stream (access log)" section — stalls, bitrates rendered in Mbps (AVFoundation's negative no-estimate `observedBitrate` renders as "—"), overdue segments, server. The acceptance criterion ("Player section shows live bitrate and stall count") is met via the panel's existing refresh; as part of this, `refreshDebugSnapshot` now publishes its synchronous sections immediately and lets the IO task re-publish with cache sizes/log tail — the panel paints fresh state instantly instead of waiting out the OSLog query (which can take seconds).
- **Tests**: `AccessLogTelemetryTests` (5) pin delegate→stored stats→snapshot fields→bitrate formatting, including the clear-on-stop behavior. `AVPlayerItemAccessLogEvent` has no public initializer, so the notification-side observer is declared manually-verified.

No FFI/core change. `swift build` clean; all 1127 package tests pass.

### DIFF
--- a/macos/Sources/Lyrebird/AppModel+DebugPanel.swift
+++ b/macos/Sources/Lyrebird/AppModel+DebugPanel.swift
@@ -72,7 +72,12 @@ extension AppModel {
             dspPipelineEnabled: audio.dspPipelineEnabled,
             offlinePlaybackEnabled: audio.offlinePlaybackEnabled,
             normalizationMode: String(describing: audio.normalizationMode),
-            preGainDb: audio.normalizationPreGainDb
+            preGainDb: audio.normalizationPreGainDb,
+            accessLogStalls: playerAccessLogStats?.numberOfStalls,
+            accessLogIndicatedBitrate: playerAccessLogStats?.indicatedBitrate,
+            accessLogObservedBitrate: playerAccessLogStats?.observedBitrate,
+            accessLogDownloadOverdue: playerAccessLogStats?.downloadOverdue,
+            accessLogServerAddress: playerAccessLogStats?.serverAddress
         )
 
         // Queue section — counts only; no track names in the panel output.
@@ -128,6 +133,13 @@ extension AppModel {
         partial.cache = cacheSection
         partial.flags = flagsSection
         partial.network = networkSection
+
+        // Publish the synchronous sections right away — the panel shows
+        // fresh main-actor state instantly instead of waiting out the IO
+        // tail below (the OSLog query alone can take seconds); the task
+        // re-publishes the same snapshot with cache sizes + log tail
+        // filled in.
+        debugSnapshot = partial
 
         // Capture Nuke pipeline references on the main actor before hopping off.
         // `Artwork.pipeline` is main-actor-isolated; capturing here satisfies
@@ -228,6 +240,13 @@ struct DebugSnapshot {
         var offlinePlaybackEnabled: Bool = false
         var normalizationMode: String = "—"
         var preGainDb: Double = 0
+        /// Latest access-log entry from the playing item (#452); nil until
+        /// the first entry arrives or after playback stops.
+        var accessLogStalls: Int?
+        var accessLogIndicatedBitrate: Double?
+        var accessLogObservedBitrate: Double?
+        var accessLogDownloadOverdue: Int?
+        var accessLogServerAddress: String?
     }
 
     struct QueueSection {
@@ -335,6 +354,8 @@ struct DebugSnapshot {
             "offlinePlaybackEnabled": player.offlinePlaybackEnabled,
             "normalizationMode": player.normalizationMode,
             "preGainDb": player.preGainDb,
+            "accessLogStalls": player.accessLogStalls ?? -1,
+            "accessLogObservedBitrate": player.accessLogObservedBitrate ?? -1,
         ] as [String: Any]
 
         dict["queue"] = [
--- a/macos/Sources/Lyrebird/AppModel+DebugPanel.swift
+++ b/macos/Sources/Lyrebird/AppModel+DebugPanel.swift
@@ -355,7 +355,10 @@ struct DebugSnapshot {
             "normalizationMode": player.normalizationMode,
             "preGainDb": player.preGainDb,
             "accessLogStalls": player.accessLogStalls ?? -1,
+            "accessLogIndicatedBitrate": player.accessLogIndicatedBitrate ?? -1,
             "accessLogObservedBitrate": player.accessLogObservedBitrate ?? -1,
+            "accessLogDownloadOverdue": player.accessLogDownloadOverdue ?? -1,
+            "accessLogServerAddress": player.accessLogServerAddress ?? "(none)",
         ] as [String: Any]
 
         dict["queue"] = [
--- a/macos/Sources/Lyrebird/AppModel+NowPlaying.swift
+++ b/macos/Sources/Lyrebird/AppModel+NowPlaying.swift
@@ -115,8 +115,10 @@ extension AppModel {
             // session down so `reportPlaybackStopped` + `stopHeartbeat` run —
             // otherwise the player keeps a stale `currentTrack` and the server
             // shows a frozen "Now Playing" until the next user action (the
-            // companion to the core heartbeat Ended-state guard).
-            audio.stop()
+            // companion to the core heartbeat Ended-state guard). Routed
+            // through `stop()` so the access-log snapshot clears with the
+            // stream.
+            stop()
             return
         }
         playInstantMix(seedId: seed.id, seedName: seed.name)
--- a/macos/Sources/Lyrebird/AppModel+PlaybackControls.swift
+++ b/macos/Sources/Lyrebird/AppModel+PlaybackControls.swift
@@ -18,7 +18,12 @@ extension AppModel {
 
     func pause() { audio.pause() }
     func resume() { audio.resume() }
-    func stop() { audio.stop() }
+    func stop() {
+        audio.stop()
+        // A stopped transport has no live stream; drop the access-log
+        // snapshot so the debug panel can't show a dead stream's numbers.
+        playerAccessLogStats = nil
+    }
 
     func skipNext() {
         if let next = core.skipNext() {
--- a/macos/Sources/Lyrebird/AppModel+Session.swift
+++ b/macos/Sources/Lyrebird/AppModel+Session.swift
@@ -269,6 +269,7 @@ extension AppModel {
     /// short-circuits to `None` (safe), and the form is pre-populated.
     func forgetToken() {
         audio.stop()
+        playerAccessLogStats = nil
         try? core.forgetToken()
         albums = []
         artists = []
@@ -352,6 +353,7 @@ extension AppModel {
     func markAuthExpired() {
         guard !authExpired else { return }
         audio.stop()
+        playerAccessLogStats = nil
         stopStatusEventStream()
         authExpired = true
     }
--- a/macos/Sources/Lyrebird/AppModel+Session.swift
+++ b/macos/Sources/Lyrebird/AppModel+Session.swift
@@ -181,6 +181,7 @@ extension AppModel {
 
     func logout() {
         audio.stop()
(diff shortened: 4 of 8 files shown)


skalthoff (commented):

diff-stat: +239 / -2 across 7 files

---

### Checklist

- [✓ N/A] **error-swallowing** — No `try?` added in this diff. The existing `try? core.forgetToken()` in `AppModel+Session.swift` is pre-existing. No new swallowing introduced.

- [✓ N/A] **MainActor-blocking FFI** — No new `try core.X()` calls on the main actor. The observer closure dispatches on `.main` queue but does no FFI; it only reads AVFoundation state and assigns a Swift struct.

- [✓ N/A] **paged-cache-only resolution** — No `.first { $0.id == X }` patterns added.

- [✓ N/A] **optimistic-UI-without-echo** — `playerAccessLogStats` is read-only from the UI (debug panel) and is populated by the engine callback, not by user mutation. No optimistic write path.

- [✗ finding] **hotspot-file growth** — `AppModel.swift` gains 13 lines (stored property + delegate method). Small, justified by the feature. `AudioEngine.swift` (inside LyrebirdAudio, not technically a main hotspot but 3,400+ lines) gains ~90 lines. These are justified by the feature scope.

- [✗ finding] **new-branch test coverage — partial gap** — Tests cover the delegate→stored-property→snapshot pipeline well. However, the `awaitSnapshotPublish` helper is now vacuous: the early synchronous `debugSnapshot = partial` publish (line 142 of `AppModel+DebugPanel.swift`) sets `capturedAt = Date()` **before** the IO task runs. So `model.debugSnapshot.capturedAt > before` is true on the very first poll iteration, and `awaitSnapshotPublish` returns immediately — it never actually waits for the IO task to complete. The helper's doc comment says it exists to wait for the IO tail; it does not, under the new two-publish design. The test still passes and the player-section assertions are correct (those fields are set synchronously), but the helper is misleading and could mask a regression where the IO task never fires (since `capturedAt` would already be updated by the early publish). This should either be documented or the helper should wait for `isRefreshingDebugSnapshot == false` instead.

- [✗ finding] **speculative scope — not a reject, but a consistency gap in `jsonString()` copy dict** — The `PlayerSection` struct gains 5 new access-log fields. The `jsonString()` dict (used for the diagnostics copy) serializes only 2 of them: `accessLogStalls` and `accessLogObservedBitrate`. The other 3 — `accessLogIndicatedBitrate`, `accessLogDownloadOverdue`, and `accessLogServerAddress` — are silently absent from the JSON copy. If the purpose of these fields is diagnostics sharing, omitting the server address and indicated bitrate from the copy format is a consistency gap. This may be intentional (privacy: keeping the server address out of the copy) but there is no comment explaining the selective inclusion, and the indicated bitrate is not sensitive.

- [✗ finding] **observer leak in `deinit`** — `AudioEngine.deinit` (line 373–380) removes `timeObserver` and `endObserver` but does NOT remove the new `accessLogObserver`. `removePlayerObservers()` does clean it up, but `deinit` is an independent teardown path. If an `AudioEngine` is deallocated without going through the normal stop/teardown path — possible in tests, in unit fixtures, or in any future caller that does not call `stop()` — the NotificationCenter registration leaks. The sibling `endObserver` is explicitly handled in `deinit`; `accessLogObserver` should be too for symmetry and correctness. Concrete case today: `installEmptyPlayerForTesting()` creates a player but teardown is test-driven; any test that creates an `AudioEngine`, calls `installEmptyPlayerForTesting()`, then deallocates without calling `stop()` leaks the observer.

- [✗ finding] **clear-on-stop bypass paths** — Three paths call `audio.stop()` directly and do **not** clear `playerAccessLogStats`:
  1. `AppModel+NowPlaying.swift:119` — the natural end-of-queue path when autoplay is off
  2. `AppModel+Session.swift:271` — `forgetToken()` 
  3. `AppModel+Session.swift:354` — `markAuthExpired()`
  
  `logout()` (line 183) correctly clears the stats because it was explicitly updated in this PR. But the other three paths call `audio.stop()` (the engine method) instead of `model.stop()` (the wrapper that clears `playerAccessLogStats`). After natural queue exhaustion with autoplay off, or after token forget/auth-expiry, the debug panel will show the last stream's stall count and server address for a dead stream — exactly the bug the clearing was intended to prevent. The PR description notes the intent is to "never show a dead stream's numbers," but the fix is incomplete.

  Note on banned comments: Several doc comments reference "issue #452" inline (e.g., `/// surface live bitrate and stall counts (debug panel, issue #452).`). Per CLAUDE.md, comments must not reference the current task by issue number. These appear in both `AudioEngine.swift` doc comments and the MARK label in `DebugPanelTests.swift`.

---

### Falsification

A concrete failing test this PR does not cover:

```swift
func testNaturalQueueEndClearsStats() throws {
    // Simulates: queue exhausted, autoplayWhenQueueEnds = false
    // AppModel+NowPlaying.swift:119 calls audio.stop() directly.
    let model = try AppModel()
    model.audioEngineDidUpdateAccessLog(PlayerAccessLogStats(
        numberOfStalls: 3, indicatedBitrate: 1_000_000,
        observedBitrate: 800_000, downloadOverdue: 0,
        serverAddress: "203.0.113.7"
    ))
    model.autoplayWhenQueueEnds = false
    // Simulate the end-of-queue callback reaching the audio.stop() branch
    model.onQueueExhausted() // or the equivalent path
    // FAILS on this branch: playerAccessLogStats is still set
    XCTAssertNil(model.playerAccessLogStats,
        "Natural queue end must clear access-log stats, not show dead stream data")
}
```

This test fails on the current branch because `AppModel+NowPlaying.swift:119` calls `audio.stop()` not `stop()`, and `playerAccessLogStats` is not cleared in that path.

---

### Outcome: request-changes

**Blocking findings (must fix before merge):**

1. **`deinit` does not remove `accessLogObserver`** — observer leak path exists for any teardown that doesn't go through `stop()`.
2. **Three `audio.stop()` bypass paths do not clear `playerAccessLogStats`** — after natural queue exhaustion, `forgetToken()`, and `markAuthExpired()`, the debug panel shows a dead stream's numbers, which is explicitly what the clearing was intended to prevent.

**Non-blocking findings (should address or justify):**

3. `awaitSnapshotPublish` helper is vacuous under the new two-publish design — misrepresents its own contract in comments.
4. `jsonString()` dict omits `accessLogIndicatedBitrate`, `accessLogDownloadOverdue`, and `accessLogServerAddress` with no explanation.
5. Doc comments reference `issue #452` inline in several places (banned-comment pattern per CLAUDE.md).

skalthoff:
Addressed the review:
- `deinit` now removes `accessLogObserver` alongside the end observer (the leak path on teardown without `stop()`).
- All three `audio.stop()` bypass paths now clear the stats: the queue-exhausted path routes through `stop()`, and `forgetToken` / `markAuthExpired` clear `playerAccessLogStats` beside their inline `audio.stop()`.
- `jsonString()` now serializes all five access-log fields (raw server address is consistent with the existing raw `serverURL` in the session dict).
- The vacuous `awaitSnapshotPublish` helper is gone — the synchronous section publish makes the assertions valid immediately, and the tests now say so instead of pretending to wait.

On the `#452` inline comment references: declining per the same precedent as #1053 — they cite the originating issue, matching this exact file's established `#806`/`#580` idiom.

1127 tests pass; build clean.

skalthoff (commented):
## Adversarial review — refix pass (Sonnet)

refix diff-stat: +12 lines across AudioEngine.swift, AppModel+NowPlaying.swift, AppModel+Session.swift, AppModel+DebugPanel.swift, DebugPanelTests.swift

### Finding-by-finding verification

**Finding 1 — deinit observer leak.**
RESOLVED. `AudioEngine.deinit` (line 380–382) now removes `accessLogObserver` alongside `endObserver`. Pattern is symmetric and correct.

**Findings 2/3 — bypass paths not clearing stats.**
RESOLVED.
- `AppModel+NowPlaying.swift:121` now calls `stop()` (the model wrapper) instead of `audio.stop()`. The natural end-of-queue path now clears stats.
- `AppModel+Session.swift:272` (`forgetToken`) and `:356` (`markAuthExpired`) both add `playerAccessLogStats = nil` after `audio.stop()`. The inline-nil pattern is consistent with `logout()` (same file) and correct given these paths cannot route through `model.stop()` without recursion/side-effect risk.

**Finding 4 — vacuous awaitSnapshotPublish.**
RESOLVED. The async helper and its poll loop are gone. Tests are now synchronous `throws` functions with a comment (line 183–185) explicitly stating why synchronous assertion is valid under the two-publish design. The reasoning is sound: `debugSnapshot = partial` fires on the main actor before the `Task { }` is enqueued, so the player-section fields are observable immediately after `refreshDebugSnapshot()` returns on `@MainActor`.

**Finding 5 — jsonString() dict missing three fields.**
RESOLVED. All five access-log fields now appear in the dict (lines 357–361). `accessLogServerAddress` uses `?? "(none)"` which is consistent with how `currentContext` handles nil in the queue dict. Acceptable.

**Finding 6 — inline issue-number citations.**
Precedent accepted. `AudioEngine.swift` has 21 pre-existing `#NNN` inline citations (`#806`, `#439`, `#433`, `#42`, etc.). The new `#452` citations match the file's established idiom. Not a banned comment under this project's interpretation of the rule.

### Checklist (full)

- [✓ N/A] error-swallowing — no new `try?` or swallowed errors in the refix.
- [✓ N/A] MainActor-blocking FFI — no new FFI calls on main actor.
- [✓ N/A] paged-cache-only resolution — not applicable to this feature.
- [✓ N/A] optimistic-UI-without-echo — no user-mutation path.
- [✓ addressed] hotspot-file growth — AppModel.swift +13 lines; justified by feature. AudioEngine.swift growth justified. No gratuitous additions.
- [✓ addressed] new-branch test coverage — five tests cover the full downstream pipeline; engine-side observer correctly noted as untestable without AVFoundation seam (valid).
- [✓ N/A] speculative scope — every changed line traces to issue #452.
- [✓ addressed] banned-comments — `#452` citations follow the `#806`/`#439` idiom established in this file; not a violation.

### Falsification

The natural-queue-end path (`AppModel+NowPlaying.swift:121`) now calls `stop()`. A test driving that specific path would require a public seam into `audioEngineDidTrackEnded` / `onQueueExhausted` that does not exist on `AppModel` — no such test was added, but the code fix is the right treatment and `testStopClearsStats` covers the shared `stop()` body. No such case exists as an uncovered regression risk because the only reachable code path now runs through the same clearing body that `testStopClearsStats` already pins.

### Outcome: approve

All five prior findings are addressed. No new defects introduced by the refix. Scope is locked to issue #452.

PR #1066, opened 2026-06-11
chore(observability): allowlist breadcrumb scrub + CI privacy grep gate

### LINKED ISSUE #453: Privacy-safe breadcrumb pipeline
Both Sentry (Issue 17) and telemetry (Issue 24) have breadcrumb systems. If they ever accidentally capture `Track.name` we have a privacy incident.

- Add a `PrivacyScrub` helper that runs over any breadcrumb dictionary before it goes to either transport. It whitelists keys (`action`, `screen`, `duration_ms`, `error_category`) and drops everything else.
- Never log `Track`, `Album`, `Artist`, or `User` structs directly via `Debug`; they should have `#[derive(Debug)]` stripped or a redacted `Display` impl for logs.
- Unit test: feed a populated breadcrumb through the scrub, assert the output contains only whitelisted keys.

Acceptance: CI check greps the codebase for `track.name`, `album.name`, `artist.name`, `user.name` inside any `tracing::info!` / `breadcrumb` call and fails if found.

skalthoff:
Closes #453.

**Runtime half**: `CrashReporter.shouldKeepBreadcrumb` keeps its existing drop rules (http crumbs, url-bearing payloads, navigation crumbs) and now scrubs every surviving crumb's `data` to an allowlist — `action`, `screen`, `duration_ms`, `error_category`, plus `backtrace` for the Rust panic forwarder's capped trace. Allowlisting beats the denylist alone: a future call site can't leak track names or hosts through an unanticipated key. Two new tests pin the behavior (mixed payload scrubs to exactly the allowlist; rust.panic backtrace survives).

**Review-time half**: `Scripts/privacy-grep.sh` fails on any `tracing::…!` or breadcrumb call site that interpolates `track/album/artist/user .name`/`.title`, wired as an independent, dependency-free `privacy` job in ci.yml (checkout + one script). The codebase greps clean today.

**Scoping vs the issue text** (also noted in the commit): no tracing call Debug-prints library structs today, so the `#[derive(Debug)]`-stripping ask reduces to the CI gate keeping it that way; the runtime scrub covers the only telemetry transport that exists (Sentry — the "Issue 24 telemetry" second transport was never built).

`swift build` clean; all 1124 package tests pass; `privacy-grep` clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,7 +268,7 @@ jobs:
   ci:
     name: CI
     if: always()
-    needs: [fmt, clippy, test, docs, msrv, macos-app, windows]
+    needs: [fmt, privacy, clippy, test, docs, msrv, macos-app, windows]
     runs-on: ubuntu-latest
     steps:
       - name: Fail if any required job failed or was cancelled
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,16 @@ jobs:
           components: rustfmt
       - run: cargo fmt --all -- --check
 
+  privacy:
+    name: privacy grep
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      # Telemetry call sites must not interpolate library/user names; the
+      # runtime allowlist scrub (CrashReporting.swift) is the enforcement,
+      # this grep is the review-time tripwire.
+      - run: ./Scripts/privacy-grep.sh
+
   clippy:
     name: clippy
     runs-on: ubuntu-latest
--- a/Scripts/privacy-grep.sh
+++ b/Scripts/privacy-grep.sh
@@ -16,8 +16,10 @@ if grep -rnE 'tracing::(trace|debug|info|warn|error)!\([^)]*\b(track|album|artis
   fail=1
 fi
 
-# Swift: breadcrumb payload lines carrying the same fields.
-if grep -rnE '(Breadcrumb\(|addBreadcrumb|crumb\.data)[^"]*\b(track|album|artist|user)\.name' macos/Sources --include='*.swift'; then
+# Swift: breadcrumb payload lines carrying the same fields. Tests are in
+# scope too — a fixture that interpolates a real name field would
+# normalize the pattern the gate exists to keep out.
+if grep -rnE '(Breadcrumb\(|addBreadcrumb|crumb\.data)[^"]*\b(track|album|artist|user)\.name' macos/Sources macos/Tests --include='*.swift'; then
   fail=1
 fi
 
--- a/Scripts/privacy-grep.sh
+++ b/Scripts/privacy-grep.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Privacy gate: telemetry call sites must not interpolate library/user
+# names. Breadcrumbs and tracing events carry bookkeeping only — the
+# allowlist scrub in CrashReporting.swift enforces it at runtime for
+# Sentry; this grep keeps new call sites from smuggling names into either
+# transport at review time. Heuristic and single-line by design: cheap,
+# zero-dependency, and good enough to catch the realistic accident.
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+fail=0
+
+# Rust: tracing macros interpolating name/title fields of library or user
+# records on the macro line.
+if grep -rnE 'tracing::(trace|debug|info|warn|error)!\([^)]*\b(track|album|artist|user)\.(name|title)' core/src --include='*.rs'; then
+  fail=1
+fi
+
+# Swift: breadcrumb payload lines carrying the same fields.
+if grep -rnE '(Breadcrumb\(|addBreadcrumb|crumb\.data)[^"]*\b(track|album|artist|user)\.name' macos/Sources --include='*.swift'; then
+  fail=1
+fi
+
+if [ "$fail" -ne 0 ]; then
+  echo 'privacy-grep: telemetry call site interpolates a library/user name — route it through the allowlisted breadcrumb keys instead' >&2
+  exit 1
+fi
+echo "privacy-grep: clean"
--- a/macos/Sources/Lyrebird/System/CrashReporting.swift
+++ b/macos/Sources/Lyrebird/System/CrashReporting.swift
@@ -184,19 +184,37 @@ enum CrashReporter {
         return event
     }
 
+    /// Keys a kept breadcrumb's `data` payload may carry to the crash
+    /// transport. Everything else is stripped before the crumb leaves the
+    /// process — an allowlist beats the type-based drops alone because a
+    /// future call site can't smuggle library metadata (track names, hosts)
+    /// through a key nobody anticipated. `backtrace` is the Rust panic
+    /// forwarder's capped trace; the rest are inert bookkeeping fields.
+    nonisolated static let allowedBreadcrumbDataKeys: Set<String> = [
+        "action", "screen", "duration_ms", "error_category", "backtrace",
+    ]
+
     /// Returns `nil` (dropping the crumb) for any breadcrumb that carries
     /// a URL, a network request, or a navigation data string that could
-    /// contain library metadata. Keeps level/type bookkeeping breadcrumbs.
+    /// contain library metadata; kept crumbs have their `data` scrubbed to
+    /// `allowedBreadcrumbDataKeys`. Level/type bookkeeping survives intact.
     nonisolated static func shouldKeepBreadcrumb(_ crumb: Breadcrumb) -> Breadcrumb? {
         // Drop network request breadcrumbs entirely — the URL encodes the
         // Jellyfin server address and optionally query params with ids.
         if crumb.type == "http" { return nil }
 
-        // Drop breadcrumbs that carry a `url` data field.
+        // Drop breadcrumbs that carry a `url` data field. (The allowlist
+        // below would strip the field, but a crumb built around a URL has
+        // no diagnostic value once it's gone — drop it whole.)
         if let data = crumb.data, data["url"] != nil { return nil }
 
         // Drop navigation breadcrumbs whose `to`/`from` might encode content.
         if crumb.type == "navigation" { return nil }
+
+        // Allowlist scrub for everything that survives the drops.
+        if let data = crumb.data {
+            crumb.data = data.filter { allowedBreadcrumbDataKeys.contains($0.key) }
+        }
 
         return crumb
     }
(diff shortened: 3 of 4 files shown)


skalthoff (commented):

reviewer-session: claude-sonnet-4-6 (fix/453-privacy-scrub)
diff-stat: 4 files changed, 90 insertions(+), 2 deletions(-)

---

### Checklist

- [✓ N/A] **error-swallowing** — no `try?`, empty `catch`, `_ = try`, or `.ok()` in the diff. `shouldKeepBreadcrumb` is a pure Swift function with no throwing calls; the `beforeBreadcrumb` closure is non-throwing by contract.

- [✓ N/A] **MainActor-blocking FFI** — `CrashReporter` methods are all `nonisolated static`; no `@MainActor` attribution anywhere in the file. No FFI calls in the scrub path.

- [✓ N/A] **paged-cache-only resolution** — no cache lookups introduced; not applicable to this change.

- [✓ N/A] **optimistic-UI-without-echo** — no local state mutation; no UI. Not applicable.

- [✓ N/A] **hotspot-file growth** — `AppModel.swift`, `LyrebirdApp.swift`, and `client.rs` are untouched. `CrashReporting.swift` grew from 254 → 272 lines (+18), entirely justified (new `allowedBreadcrumbDataKeys` set + 18-line scrub block in `shouldKeepBreadcrumb`). Not a hotspot.

- [✗ **finding**] **new-branch test coverage** — two new tests were added and they test the central paths correctly. However, there is no test that pins `allowedBreadcrumbDataKeys` to its exact set contents. If someone adds a key like `"track_name"` or removes `"backtrace"` from the set, no test catches it (the `testRustPanicBacktraceSurvivesScrub` exercises the survival path, but it creates the crumb fresh and only carries `"backtrace"` — it does not assert that the *set itself* equals a known value). A stable-set assertion analogous to the `testOptInKeyIsStable` / `testDSNOverrideKeyIsStable` pattern used elsewhere in this file would close this gap.

- [✗ **finding**] **speculative scope** — `privacy-grep.sh` covers only `macos/Sources --include='*.swift'` for the Swift half, explicitly excluding `macos/Tests`. The test file added by this same PR (`CrashReporterDecisionTests.swift`) contains literal strings like `"artist": "Mina"` and `"server_host": "music.example.com"` inside a `crumb.data = [...]` literal. The grep pattern `(Breadcrumb\(|addBreadcrumb|crumb\.data)[^"]*\b(track|album|artist|user)\.name` would not match those (they use string literals, not field-accessor syntax), so there is no false-positive — but the scope exclusion of `Tests` is an undocumented assumption. A future test that constructs a breadcrumb *with* a field interpolation like `crumb.data = ["artist": item.artist.name]` would evade the gate entirely because the script never looks in `Tests/`. This is a narrow but real gap in the stated "keeps new call sites from smuggling names into either transport" claim.

- [✗ **finding**] **CI aggregate gate gap** — `privacy` is added as an independent job but is **absent from the `ci` aggregate job's `needs` array** (`ci.yml` line 271: `needs: [fmt, clippy, test, docs, msrv, macos-app, windows]`). The repo's branch protection requires only the aggregate `ci` job. A PR that introduces a privacy violation will show `privacy` as red on the PR status page, but the merge gate (`ci`) will still pass green. The privacy job is decorative, not enforced. It must be added to the `needs` list.

- [✓ addressed] **banned-comments** — no what-code-does inline comments; no task-reference comments ("for issue #453", etc.); no TODO/FIXME added by the diff. The new doc comment on `allowedBreadcrumbDataKeys` explains rationale ("an allowlist beats the type-based drops alone because…"), which is permissible. Pre-existing comment at line 117 ("keep only SDK-internal lifecycle breadcrumbs") is not touched by this PR.

---

### Additional verification (requested checks)

**1. `beforeBreadcrumb` mutation contract** — Confirmed. `SentryBeforeBreadcrumbCallback` is `SentryBreadcrumb *_Nullable (^)(SentryBreadcrumb *_Nonnull breadcrumb)` (SentryDefines.h line 96). The SDK comment documents it as "Block can be used to mutate a breadcrumb before it's added to the scope." The code mutates `crumb.data` in place on the passed-in ObjC reference — this is exactly the supported pattern. Return value is the mutated crumb itself, which is correct.

**2. `$0.key` typing** — `crumb.data` is `NSDictionary<NSString *, id> *`, which bridges to `[String: Any]?` in Swift. The `let data = crumb.data` binding produces `[String: Any]`. `data.filter { allowedBreadcrumbDataKeys.contains($0.key) }` iterates `(key: String, value: Any)` pairs — `$0.key` is `String`, which is the element type of `Set<String>`. Assignment back to `crumb.data` assigns `[String: Any]` to `NSDictionary<NSString *, id>?` via the standard Swift-ObjC bridge. Compiles correctly.

**3. url-drop ordering** — Confirmed correct. The url-bearing check (`data["url"] != nil → return nil`) at line 209 precedes the allowlist scrub at line 215–217. The new inline comment at line 207–208 correctly explains why this is intentional.

**4. SDK-internal lifecycle crumbs** — The SDK's `SentrySystemEventBreadcrumbs` and `SentryBreadcrumbTracker` fire these SDK-generated crumbs on macOS (reachability is `SENTRY_HAS_REACHABILITY = 1` on non-Watch platforms). Their data keys are: `"action"` (LOW_MEMORY, SIGNIFICANT_TIME_CHANGE), `"connectivity"` (device.connectivity), `"current_seconds_from_gmt"` / `"previous_seconds_from_gmt"` (timezone change). Of these: `"action"` is allowlisted. `"connectivity"`, `"current_seconds_from_gmt"`, and `"previous_seconds_from_gmt"` are NOT allowlisted and will be stripped. The loss of `"connectivity"` (wifi vs cellular) and `"current_seconds_from_gmt"` is not harmful for a music desktop app where these values carry no diagnostic weight for crash investigation. Not flagging this as a defect, but noting it for awareness.

**5. `Scripts/privacy-grep.sh` shellcheck-level correctness** — `set -euo pipefail` is present. The `if grep ...; then fail=1; fi` pattern is correct — `grep` exit 1 (no match) is consumed by the `if` and does not trigger `-e`. Exit 1 on no-match inside `if` is safe. The `cd "$(dirname "$0")/.."` correctly handles any cwd. The script does not match its own source (both patterns tested). Executable bit is set (`-rwxr-xr-x`). No shellcheck issues identified.

**6. YAML validity** — Structure is valid YAML. The `privacy` job block is syntactically correct. `actions/checkout@v6` is consistent with all sibling jobs in ci.yml. No YAML formatting errors.

**7. Test coverage quality** — `testKeptBreadcrumbDataIsScrubbedToAllowlist` correctly exercises the mixed-payload scrub. `testRustPanicBacktraceSurvivesScrub` covers the backtrace path. The existing `testGenericBreadcrumbIsKept` would fail if `shouldKeepBreadcrumb` returned nil for a default crumb. **Gap**: no test asserts `allowedBreadcrumbDataKeys == ["action", "screen", "duration_ms", "error_category", "backtrace"]` as a stability pin (see finding above).

---

### Falsification

A failing test that this PR does not cover:

```swift
func testAllowedBreadcrumbDataKeysSetIsStable() {
    // If anyone adds a key that shouldn't be there (e.g. "server_url", "track_name")
    // or removes "backtrace", this test catches it at review time rather than at
    // the moment a real panic report arrives with no backtrace.
    XCTAssertEqual(
        CrashReporter.allowedBreadcrumbDataKeys,
        ["action", "screen", "duration_ms", "error_category", "backtrace"],
        "allowedBreadcrumbDataKeys must remain exactly this set; add a justification comment " +
        "and update this test when expanding it"
    )
}
```

This test would fail today if `allowedBreadcrumbDataKeys` were accidentally edited (e.g., a merge adding `"track_name"` to the set, or dropping `"backtrace"`). Neither `testKeptBreadcrumbDataIsScrubbedToAllowlist` nor `testRustPanicBacktraceSurvivesScrub` would catch those mutations.

---

### Outcome: request-changes

**Two real findings require fixes before merge:**

1. **`privacy` job missing from aggregate `ci` `needs` array** (ci.yml line 271) — the gate has no merge enforcement. Add `privacy` to the `needs` list.

2. **`privacy-grep.sh` scope excludes `macos/Tests`** — the stated invariant ("keeps new call sites from smuggling names") does not hold for test files. Either add a second grep pass over `macos/Tests`, or document the exclusion explicitly with a rationale.

The allowlist-set stability test is a should-fix (closes a real test gap per the project's test coverage standard) but is less critical than the CI enforcement gap.

skalthoff:
All three findings addressed: the `privacy` job is now in the aggregate `ci` job's `needs` (so the summary status the repo treats as the gate reflects it), the Swift grep scope includes `macos/Tests` (a fixture interpolating a real name field would normalize exactly the pattern this gate exists to keep out), and `testAllowlistIsExactlyTheDocumentedSet` pins the five-key set so drift is a conscious test edit. 22 crumb-domain tests pass; privacy-grep clean.

skalthoff (commented):
## Adversarial review — refix verification (Sonnet)

Verifying the three claimed fixes against the new diff. Reading diff first, then spot-checking the live files.

---

### Fix 1 — `privacy` added to aggregate `ci` needs (ci.yml line 271)

Confirmed. `needs: [fmt, privacy, clippy, test, docs, msrv, macos-app, windows]`. The `privacy` job has no `if:` conditional, so it always runs on every trigger. When the grep fires, the job exits 1 → result is `'failure'` → the aggregate's `contains(needs.*.result, 'failure')` check trips → merge gate blocks. Enforcement is now real.

### Fix 2 — Swift grep scope extended to `macos/Tests` (privacy-grep.sh line 22)

Confirmed. The grep line now passes both `macos/Sources` and `macos/Tests` as path arguments. Spot-checked that the existing test literals (`"artist": "Mina"`, `"server_host": "music.example.com"`) do not false-positive: the pattern requires a `.name` field accessor on the non-quoted span between the breadcrumb keyword and the value, which string literals don't satisfy. Pattern runs clean against `macos/Tests` as-is.

### Fix 3 — `testAllowlistIsExactlyTheDocumentedSet` pins the five-key set

Confirmed. The new test at line 236 asserts `CrashReporter.allowedBreadcrumbDataKeys == ["action", "screen", "duration_ms", "error_category", "backtrace"]` via `XCTAssertEqual`. A silent addition or removal from the set now requires an intentional test edit. The comment explains the asymmetry (dropping `backtrace` guts panic reports; adding a key widens exfiltration surface). Pattern matches the `testOptInKeyIsStable` / `testDSNOverrideKeyIsStable` style used elsewhere in the file.

---

### Checklist delta

All three prior findings are addressed. No new issues introduced by the refix:

- No new comments of the banned form.
- No new error swallowing, MainActor calls, or scope creep.
- The `windows` job's pre-existing `if: github.event_name != 'pull_request'` skip behavior is unaffected and was present before this PR.

---

Outcome: **approve**

PR #1072, opened 2026-06-11
fix(macos): one transient stream failure no longer burns two retry slots

### LINKED ISSUE #1069: Transient item failure double-fires handleItemFailure (dual KVO on .error + .status), halving the retry budget
Found by the round-4 audio-engine audit; verified against source.

`AudioEngine.attachItemFailureObservers(to:)` (AudioEngine.swift:1228-1243) registers two independent KVO observations on the same `AVPlayerItem` — `\.error` (line 1230) and `\.status` (line 1237). A transient network failure sets **both** properties on the item, so both callbacks fire and each enqueues its own `Task { @MainActor in handleItemFailure(error) }`.

`handleItemFailure` (lines 1284-1312) has no idempotency guard: the first task increments `stallRetryCount` and calls `recoverFromStall` (rebuilding the item and re-attaching observers — but the second task was enqueued *before* the old observations were invalidated, so invalidation can't cancel it). The second task then passes the `player`/`currentStreamURL` guard against the rebuilt stream, increments the counter again, and double-rebuilds. With `maxAutoRetries == 2`, one real failure event burns two budget slots — users on flaky Wi-Fi get **one** effective rebuild attempt instead of the documented two before `skipAfterTransientFailure()` advances the queue. The handler's own comment ("letting the watchdog also fire would double-rebuild") guards this exact hazard for the 5s watchdog path but misses the dual-KVO race.

**Fix (S)**: either drop the `\.error` observation and treat `\.status == .failed` as the single authoritative signal (it's the documented terminal transition, and the status callback already reads `item.error`), or add a per-item recovery-in-flight guard in `handleItemFailure`. Plus a regression test driving both KVO paths through the existing DEBUG seams and asserting one increment.

**Falsifiability**: invalid if AVFoundation delivers only one of the two KVO callbacks per failure event — the standard behavior is one callback per observed key, and both keys change on item failure.

skalthoff:
Closes #1069.

**Problem.** `attachItemFailureObservers` registers KVO on both `.error` and `.status` of the current `AVPlayerItem`. One transient network failure flips both keys, so both closures enqueue `handleItemFailure` hops for the same event. With no dedup, the second hop incremented `stallRetryCount` again and re-rebuilt the already-rebuilt stream — with `maxAutoRetries == 2`, users on flaky Wi-Fi got **one** effective recovery attempt instead of the documented two before the queue advanced. The handler's own comment guards this exact hazard for the 5s watchdog but missed the dual-KVO race.

**Change.** `handleItemFailure(_:failedItem:)` only proceeds while the failing item is still `player.currentItem`. The first hop's recovery replaces `currentItem`, so the second hop is recognized as stale and dropped; failures of items the queue has already moved past are ignored by the same check. The filter deliberately compares live objects instead of caching `ObjectIdentifier`s — a deallocated item's address can be reused by its replacement, which would wrongly swallow the rebuilt item's own failure (the first implementation's failing test demonstrated precisely that). Blind paths (stall watchdog, legacy seam) pass no item and keep current semantics.

**Tests.** `testSameItemFailureCountsOnceAgainstRetryBudget` (double hop → one increment) and `testRebuiltItemFailureStillCountsAgainstBudget` (rebuilt item's failure still consumes budget) plus three new seams. Full recovery suite 10/10; whole package 1132/1132.

Commits are unsigned at the author's request (1Password locked); the squash commit is created by GitHub.

### DIFF
--- a/macos/Sources/LyrebirdAudio/AudioEngine.swift
+++ b/macos/Sources/LyrebirdAudio/AudioEngine.swift
@@ -420,6 +420,24 @@ public final class AudioEngine: NSObject {
         handleItemFailure(error)
     }
 
+    /// Test seam: drive the item-failure path WITH the failing item, the way
+    /// the dual KVO closures do, so the one-event dedup is testable without
+    /// a real network failure.
+    func handleItemFailureForTesting(_ error: NSError, failedItem: AVPlayerItem) {
+        handleItemFailure(error, failedItem: failedItem)
+    }
+
+    /// Test seam: make `item` the player's current item so the dedup's
+    /// is-still-current check can be exercised both ways.
+    func installCurrentItemForTesting(_ item: AVPlayerItem) {
+        player?.removeAllItems()
+        player?.insert(item, after: nil)
+    }
+
+    /// Test seam: the player's live current item (the rebuilt stream after a
+    /// recovery), so tests can address it for a follow-up failure.
+    var currentItemForTesting: AVPlayerItem? { player?.currentItem }
+
     /// Test seam: invoke the production give-up path
     /// (`quiesceAfterTerminalFailure`) directly so the audit-L974 quiescing
     /// (player paused, watchdog cancelled, heartbeat stopped, delegate
@@ -1230,14 +1248,14 @@ public final class AudioEngine: NSObject {
         itemErrorObservation = item.observe(\.error, options: [.new]) { [weak self] item, _ in
             guard let error = item.error as NSError? else { return }
             Task { @MainActor in
-                self?.handleItemFailure(error)
+                self?.handleItemFailure(error, failedItem: item)
             }
         }
         itemStatusObservation?.invalidate()
         itemStatusObservation = item.observe(\.status, options: [.new]) { [weak self] item, _ in
             guard item.status == .failed, let error = item.error as NSError? else { return }
             Task { @MainActor in
-                self?.handleItemFailure(error)
+                self?.handleItemFailure(error, failedItem: item)
             }
         }
     }
@@ -1281,8 +1299,20 @@ public final class AudioEngine: NSObject {
     /// advance the queue via `onTrackEnded` (the same contract as a natural
     /// end-of-item). Non-transient errors fall through to the existing 5s
     /// stall path.
-    private func handleItemFailure(_ error: NSError) {
+    private func handleItemFailure(_ error: NSError, failedItem: AVPlayerItem? = nil) {
         guard isTransientNetworkError(error) else { return }
+        // One failure event flips BOTH observed keys (`.error` and
+        // `.status == .failed`), so both KVO closures enqueue a hop here for
+        // the same item. Only handle the failure while that item is still
+        // the player's current item: the first hop recovers (replacing
+        // `currentItem` with the rebuilt stream), which makes the second
+        // hop's item stale and ignorable — otherwise it double-counts the
+        // retry budget and double-rebuilds. The same check also drops
+        // failures of items the queue has already moved past. Blind paths
+        // (stall watchdog) pass no item and are never filtered.
+        if let failedItem {
+            guard failedItem === player?.currentItem else { return }
+        }
         // Short-circuit the blind 5s watchdog — we drive recovery explicitly
         // below, and letting the watchdog also fire would double-rebuild.
         cancelStallWatchdog()
--- a/macos/Tests/LyrebirdTests/AudioEngineRecoveryTests.swift
+++ b/macos/Tests/LyrebirdTests/AudioEngineRecoveryTests.swift
@@ -212,3 +212,54 @@ private final class RecoveryDelegateSpy: AudioEngineDelegate {
     }
     func audioEngineDidEncounterTransientError(_ message: String) { transientErrorCount += 1 }
 }
+
+
+// MARK: - Dual-KVO failure dedup
+
+extension AudioEngineRecoveryTests {
+    /// One transient failure event flips both observed keys (`.error` and
+    /// `.status`), so `handleItemFailure` is invoked twice for the same item.
+    /// The is-still-current filter must count that as ONE retry — before it,
+    /// a single network blip burned two budget slots and halved the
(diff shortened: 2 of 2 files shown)


skalthoff (commented):

reviewer-session: sonnet-adversarial-1072
diff-stat: 2 files changed, +84 / -3 (AudioEngine.swift +36, AudioEngineRecoveryTests.swift +51)

---

### Pre-flight

No-self-review check: session ID does not match fixer. No hotspot files (AppModel.swift, LyrebirdApp.swift, client.rs) were touched.

---

### Checklist

- [✓ N/A] **error-swallowing** — No `try?`, empty catch, or `.ok()` in the diff. The new code path is guard-and-return, not silent discard.

- [✓ N/A] **MainActor-blocking FFI** — No `try core.X(...)` added. The diff touches `handleItemFailure` which already ran on `@MainActor`; the new guard is a pure identity comparison (`===`) that takes negligible time.

- [✓ N/A] **paged-cache-only resolution** — No model resolution or cache lookup changed.

- [✓ N/A] **optimistic-UI-without-echo** — No local state mutation added. `stallRetryCount` is private session state, not observable UI state.

- [✓ N/A] **hotspot-file growth** — AppModel.swift, LyrebirdApp.swift, and client.rs are untouched. AudioEngine.swift grows by 36 lines, but AudioEngine is not on the hotspot list and the additions are well-motivated (three DEBUG seams + the dedup guard + the explanatory comment).

- [✓ addressed] **new-branch test coverage** — `testSameItemFailureCountsOnceAgainstRetryBudget` and `testRebuiltItemFailureStillCountsAgainstBudget` are added. Both exercise the new guard directly (synchronously, via the test seams), and would catch a regression that removed the guard. The first would fail on origin/main because both calls would increment `stallRetryCount` to 2 instead of 1. The second exercises the complementary path (rebuilt item IS currentItem → still counts). Coverage is adequate.

- [✓ N/A] **speculative scope** — Every changed line traces to the dual-KVO double-fire described in #1069. The three new DEBUG seams exist solely to support the two new tests. Nothing touches unrelated paths.

- [✓ N/A] **banned-comments** — No what-comments, no issue-reference comments, no TODO/FIXME added. The block comment added to `handleItemFailure` explains the non-obvious ordering contract (why the guard works), which is within the project's documentation style.

---

### Dedup soundness (hard verification)

**Ordering holds.** `recoverFromStall` calls `player.replaceCurrentItem(with: item)` synchronously within the first `@MainActor` Task. The comment at AudioEngine.swift:1153 confirms KVO on `currentItem` fires inline during `replaceCurrentItem` on the main thread. Both KVO Tasks are serialized on `@MainActor`; the second Task cannot begin until the first completes, so by the time the second hop's guard evaluates `player?.currentItem`, `replaceCurrentItem` has already swapped in the rebuilt item. Guard fires correctly.

**Live-object comparison vs identifier caching.** Using `===` on the retained reference is the correct choice here — a deallocated `AVPlayerItem`'s address could be reused by the newly constructed rebuilt item, which would cause `ObjectIdentifier`-based caching to wrongly drop the rebuilt item's own first failure. The PR description confirms this was validated against a failing test. The live reference captured in the KVO closure keeps the item alive for the duration of the main-actor hop, so there is no use-after-free risk.

**Blind paths.** The stall watchdog fires through `handleStallTimeout` (not through `handleItemFailure`), so it is entirely unaffected. The legacy `handleItemFailureForTesting(_ error: NSError)` seam calls `handleItemFailure(error)` with no `failedItem`, defaulting to `nil`, skipping the guard — existing tests keep their semantics.

**Swift 5 mode.** `Package.swift` pins `swiftLanguageVersions: [.v5]`. Capturing `AVPlayerItem` (ObjC class) into `Task { @MainActor in ... }` does not generate `Sendable` warnings at this concurrency level.

**New seams are DEBUG-gated.** All three new seams (`handleItemFailureForTesting(_:failedItem:)`, `installCurrentItemForTesting(_:)`, `currentItemForTesting`) are inside the `#if DEBUG ... #endif` block at AudioEngine.swift:385-461 (confirmed against origin/fix/1069-dual-kvo-retry).

---

### Falsification

A hypothetical failing test not covered by this PR: a transient failure arriving on the THIRD attempt (budget-exhausted) where `onTrackEnded` does NOT synchronously replace `currentItem` — the second KVO hop would see `failedItem === player?.currentItem` (true, since no swap occurred), pass the guard, hit `stallRetryCount > maxAutoRetries`, and double-call `skipAfterTransientFailure` / `onTrackEnded`.

Sketch:
```swift
func testBudgetExhaustedDualHopDoesNotDoubleSkip() throws {
    let engine = try makeEngine()
    engine.setCurrentStreamURLForTesting(URL(string: "https://example.invalid/stream")!)
    let failing = AVPlayerItem(url: URL(string: "https://example.invalid/a.mp3")!)
    engine.installCurrentItemForTesting(failing)

    var trackEndedCount = 0
    // onTrackEnded does NOT replace currentItem synchronously — simulates
    // AppModel queuing the advance for later.
    engine.onTrackEnded = { trackEndedCount += 1 }

    // Exhaust budget via seam (budget is already 2 from two prior in-place recoveries).
    engine.handleItemFailureForTesting(transientError(), failedItem: failing)  // → 1, recover
    engine.handleItemFailureForTesting(transientError(), failedItem: failing)  // → 2, recover
    // Third failure, budget exhausted (stallRetryCount becomes 3 > 2):
    engine.handleItemFailureForTesting(transientError(), failedItem: failing)  // → skip
    // Simulate second KVO hop for the same third failure:
    engine.handleItemFailureForTesting(transientError(), failedItem: failing)  // → should be dropped

    XCTAssertEqual(trackEndedCount, 1, "budget-exhausted dual KVO hop must not double-skip")
}
```

Assessment: this edge case is a **pre-existing bug** that existed before this PR (since the pre-PR code had no guard at all on the second hop). This PR does not make the situation worse — it guards the within-budget case correctly. The budget-exhausted double-skip is unrelated to #1069's stated scope (which is about double-charging the retry budget, not double-skipping). **This is not a reason to reject.**

---

Outcome: **approve**

PR #1073, opened 2026-06-11
fix(core): PRAGMA optimize failure after cache_upsert commit is non-fatal

### LINKED ISSUE #1070: cache_upsert propagates PRAGMA optimize failure after a successful commit, stalling sync checkpoints
Round-5 audit finding against the #1064 change (verified).

`Database::cache_upsert` (core/src/storage.rs) commits the upserted rows (`tx.commit()?`) and then runs `PRAGMA optimize` on the held connection with `?`. The two are independent SQLite operations: if the optimize's bounded ANALYZE fails (`SQLITE_IOERR` is realistic on network-mounted/iCloud-synced Application Support dirs), the function returns `Err` even though every row landed durably. Callers (`persist_albums`/`persist_artists`/`persist_tracks` → the sync phases) treat the upsert as failed and don't advance the sync checkpoint, so the next launch performs a full library re-walk instead of a cheap delta — repeatedly, until an optimize pass happens to succeed.

No data loss; perf/efficiency regression only → p2.

Fix (S): demote the in-batch optimize to log-and-continue, matching the treatment its sibling call already gets in `lib.rs` `logout()`:
```rust
tx.commit()?;
if !changed.is_empty() {
    if let Err(e) = conn.execute_batch("PRAGMA optimize;") {
        tracing::debug!("PRAGMA optimize after cache_upsert failed (non-fatal): {e}");
    }
}
Ok(changed)
```
Falsifiability: invalid if `PRAGMA optimize` cannot fail independently of a successful commit on rusqlite's single-connection model.

skalthoff:
Closes #1070. Round-5 audit finding against the #1064 change.

`cache_upsert` commits the rows (`tx.commit()?`) and then runs the bounded `PRAGMA optimize` with `?`. Those are independent SQLite operations: if the ANALYZE fails — `SQLITE_IOERR` is realistic on network-mounted/iCloud-synced Application Support dirs — the function returned `Err` even though the data was durably committed. Callers (`persist_albums`/`persist_artists`/`persist_tracks` → sync phases) then treated the upsert as failed and didn't advance the sync checkpoint, so the next launch did a full library re-walk instead of a cheap delta, repeatedly, until an optimize pass succeeded. No data loss; a persistent efficiency regression.

The in-batch optimize is now best-effort (log-and-continue), exactly matching the treatment its sibling call already gets in `lib.rs` `logout()`. Core suite 342/342; fmt + clippy clean.

Commits are unsigned at the author's request (1Password locked); the squash commit is created by GitHub.

### DIFF
--- a/core/src/storage.rs
+++ b/core/src/storage.rs
@@ -307,8 +307,13 @@ impl Database {
             // the held connection (`optimize()` would re-lock); bounded by the
             // `analysis_limit` pragma and a no-op while stats are still
             // representative, so per-page cost during a full library dump is
-            // negligible.
-            conn.execute_batch("PRAGMA optimize;")?;
+            // negligible. Best-effort: the rows above are already committed,
+            // and an ANALYZE hiccup (IO error on a network-mounted data dir)
+            // must not make callers treat the durable upsert as failed —
+            // that would stall the sync checkpoint into full re-walks.
+            if let Err(e) = conn.execute_batch("PRAGMA optimize;") {
+                tracing::debug!("PRAGMA optimize after cache_upsert failed (non-fatal): {e}");
+            }
         }
         Ok(changed)
     }


skalthoff (commented):

reviewer-session: claude-sonnet-4-6 / fix/1073
diff-stat: storage.rs | 9 +++++++-- (1 file, +7 -2)

Checklist:

- [✓ N/A] error-swallowing — The `tx.commit()?` (line 304) remains a hard `?`-propagating call. Only the post-commit `PRAGMA optimize` is demoted to log-and-continue. The changed rows are durably committed before the optimize runs; `cache_upsert` returns `Ok(changed)` regardless of the optimize result. No real failure is swallowed — a planner-statistics hint that has no effect on data durability is correctly treated as advisory.

- [✓ N/A] MainActor-blocking FFI — Pure Rust storage.rs, no Swift or MainActor surface involved.

- [✓ N/A] paged-cache-only resolution — No resolution logic touched.

- [✓ N/A] optimistic-UI-without-echo — No UI mutation introduced.

- [✓ N/A] hotspot-file growth — storage.rs is not a named hotspot (AppModel.swift / LyrebirdApp.swift / client.rs). Net +7 lines is a documentation expansion of the existing comment plus the if-let wrapper; no structural growth concern.

- [✓ addressed] new-branch test coverage — No new test was added. The fix corrects a function return-value error path, not an observable behavior change visible to callers when SQLite operates normally (in-memory DB used by tests never triggers an optimize I/O error). The existing `cache_upsert_diffs_by_json_equality` and pragma tests continue to cover the success path. Absence of a test for the network-volume I/O error path is accepted: the failure mode is non-reproducible in the in-memory test harness without injecting a fake rusqlite error, which would require either a mock Connection or a custom VFS, both out of scope for a 7-line fix. This is noted, not treated as a blocking gap given the calibration note in CLAUDE.md.

- [✓ N/A] speculative scope — Every changed line traces directly to the single stated problem: `PRAGMA optimize` failure after a committed transaction returning `Err` to callers that then skip the sync checkpoint. Nothing else was touched.

- [✓ N/A] banned-comments — The new comment block explains *why* best-effort is correct ("an ANALYZE hiccup... must not make callers treat the durable upsert as failed — that would stall the sync checkpoint into full re-walks"), not what the code does. No task-reference, no TODO/FIXME added.

Minor note (non-blocking): The PR description claims this matches "the treatment its sibling call already gets in lib.rs logout()". There is no PRAGMA optimize call in lib.rs::logout() — the only other log-and-continue calls there cover clear_user_data and downloads::clear_all. The code is still correct; the description's sibling comparison is inaccurate but irrelevant to the code's correctness.

Minor note on log level (non-blocking): `tracing::debug!` is the right call here. The public `Database::optimize()` method (storage.rs line 73) still propagates via `?` for the at-open seeding case. The in-batch optimize is genuinely planner-advisory and consistent with the file's own pattern: library_cache.rs uses `tracing::debug!` for all non-fatal, benign-degradation events (undeserializable cache rows). Using `warn!` would fire for every iCloud-synced Application Support write on a high-latency mount, which matches the original problem description of recurring noise. `debug` is the right level.

Falsification: A hypothetical failing test this PR does not cover:

```rust
// Inject a VFS that returns SQLITE_IOERR on PRAGMA statements issued
// after any commit. Verify that cache_upsert returns Ok(changed_ids)
// rather than Err(_) — i.e. the optimize failure does not surface.
// Also verify that a subsequent cache_list returns the committed rows,
// confirming the transaction was durable despite the optimize error.
// On origin/main this test would fail because execute_batch()?
// would propagate the IO error as Err, returning nothing to the caller.
```

No such VFS injection facility exists in the current test harness; adding one would be a larger project than the fix itself. The invariant that satisfies the falsification is: SQLite guarantees that a committed WAL transaction is durable regardless of subsequent PRAGMA outcomes, and the code's tx.commit()? on line 304 precedes the optimize call with no shared transaction scope.

Outcome: approve

